PR #3, opened 2021-07-22
Insert values into response table from the background worker

steve-chavez:
* inserts response body as text into response table
* inserts headers as jsonb in the response table
* inserts ContentType and timedOut in response table
* removes the uuid-ossp dependency(back to bigint)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ results/
 .python-version
 venv/
 site/
+regression.*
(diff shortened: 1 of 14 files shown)


olirice:
This needs to be `req.id`
https://github.com/supabase/pg_net/pull/3/files#diff-a32c9b5f245bf14a6693ab3dd1019f75111f38bb141eb2b7f9c94473eba018baR55

PR #15, opened 2021-07-24
Handle requests errors and responses with empty body and content-type

### LINKED ISSUE #4: Handle curl failure
For unsupported protocols

steve-chavez:
Fixes #4. When doing invalid requests:

```sql
-- invalid protocol
select net.http_get('net://supabase.io');

-- invalid host name
select net.http_get('http://new.ycombinator.com');
```

We now get:

```sql
select id, url, error_msg from net.http_request_queue;

┌────┬────────────────────────────┬────────────────────────────┐
│ id │            url             │         error_msg          │
├────┼────────────────────────────┼────────────────────────────┤
│  1 │ net://supabase.io          │ Unsupported protocol       │
│  2 │ http://new.ycombinator.com │ Couldn't resolve host name │
└────┴────────────────────────────┴────────────────────────────┘
```

Additionally, also handles cases where there's no body and content-type:

```sql
select net.http_get('http://wikipedia.org');

select id, status_code, content_type, body from net.http_response;
┌────┬─────────────┬──────────────┬────────┐
│ id │ status_code │ content_type │  body  │
├────┼─────────────┼──────────────┼────────┤
│  1 │         301 │ [NULL]       │ [NULL] │
└────┴─────────────┴──────────────┴────────┘
```

### DIFF
--- a/sql/pg_net--0.1.sql
+++ b/sql/pg_net--0.1.sql
@@ -8,6 +8,7 @@ create table net.http_request_queue(
     params jsonb not null,
     headers jsonb not null,
     timeout_milliseconds int not null,
+    error_msg text,
     -- Available for delete after this date
     delete_after timestamp not null
 );
--- a/sql/pg_net--0.1.sql
+++ b/sql/pg_net--0.1.sql
@@ -8,7 +8,6 @@ create table net.http_request_queue(
     params jsonb not null,
     headers jsonb not null,
     timeout_milliseconds int not null,
-    error_msg text,
     -- Available for delete after this date
     delete_after timestamp not null
 );
@@ -23,9 +22,9 @@ create table net.http_response(
     content_type text,
     headers jsonb,
     body text,
-    timed_out bool
+    timed_out bool,
+    error_msg text
 );
-
 
 -- Blocks until an http_request is complete
 -- API: Private
--- a/src/worker.c
+++ b/src/worker.c
@@ -291,6 +291,7 @@ worker_main(Datum main_arg)
 							int argCount = 6;
 							Oid argTypes[6];
 							Datum argValues[6];
+							char nulls[6];
 							CurlData *cdata = NULL;
 							char *contentType = NULL;
 							bool timedOut = false;
@@ -299,31 +300,41 @@ worker_main(Datum main_arg)
 							curl_easy_getinfo(eh, CURLINFO_CONTENT_TYPE, &contentType);
 							curl_easy_getinfo(eh, CURLINFO_PRIVATE, &id);
 
-							argTypes[0] = INT8OID;
-							argValues[0] = Int64GetDatum(id);
-
 							elog(DEBUG1, "GET of %ld returned http status code %d\n", id, http_status_code);
 
 							cdata = hash_search(curlDataMap, &id, HASH_FIND, &isPresent);
 
+							argTypes[0] = INT8OID;
+							argValues[0] = Int64GetDatum(id);
+							nulls[0] = ' ';
+
 							argTypes[1] = INT4OID;
 							argValues[1] = Int32GetDatum(http_status_code);
+							nulls[1] = ' ';
 
 							argTypes[2] = CSTRINGOID;
 							argValues[2] = CStringGetDatum(cdata->body->data);
-
-							elog(DEBUG1, "Body is: %s\n", cdata->body->data);
+							if(cdata->body->data[0] == '\0')
+								nulls[2] = 'n';
+							else
+								nulls[2] = ' ';
 
 							argTypes[3] = JSONBOID;
 							argValues[3] = JsonbPGetDatum(JsonbValueToJsonb(pushJsonbValue(&cdata->headers, WJB_END_OBJECT, NULL)));
+							nulls[3] = ' ';
 
 							argTypes[4] = CSTRINGOID;
 							argValues[4] = CStringGetDatum(contentType);
+							if(!contentType)
+								nulls[4] = 'n';
+							else
+								nulls[4] = ' ';
 
 							argTypes[5] = BOOLOID;
 							argValues[5] = BoolGetDatum(timedOut);
+							nulls[5] = ' ';
 
-							if (SPI_execute_with_args(insert_query.data, argCount, argTypes, argValues, NULL,
+							if (SPI_execute_with_args(insert_query.data, argCount, argTypes, argValues, nulls,
 											false, 1) != SPI_OK_INSERT)
 							{
 								elog(ERROR, "SPI_exec failed: %s", insert_query.data);
--- a/src/worker.c
+++ b/src/worker.c
@@ -77,11 +77,11 @@ header_cb(void *contents, size_t size, size_t nmemb, void *userp)
 		char *tmp = pstrdup(contents);
 		JsonbValue	key, val;
 
-		elog(DEBUG2, "FullHeaderLine: %s\n", tmp);
+		elog(DEBUG1, "FullHeaderLine: %s\n", tmp);
 
 		/*The header comes as "Header-Key: val", split by whitespace and ditch the colon later*/
 		token = strtok(tmp, " ");
-		elog(DEBUG2, "Key: %s\n", token);
+		elog(DEBUG1, "Key: %s\n", token);
 
 		key.type = jbvString;
 		key.val.string.val = token;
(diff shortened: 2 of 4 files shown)


sql/pg_net--0.1.sql
```diff
@@ -8,6 +8,7 @@ create table net.http_request_queue(
     params jsonb not null,
     headers jsonb not null,
     timeout_milliseconds int not null,
+    error_msg text,
```
steve-chavez, line 11: Needed for specifying the request error message.

src/worker.c
```diff
@@ -255,64 +255,99 @@ worker_main(Datum main_arg)
...
+							argTypes[3] = JSONBOID;
+							argValues[3] = JsonbPGetDatum(JsonbValueToJsonb(pushJsonbValue(&cdata->headers, WJB_END_OBJECT, NULL)));
+							nulls[3] = ' ';
+
+							argTypes[4] = CSTRINGOID;
+							argValues[4] = CStringGetDatum(contentType);
+							if(!contentType)
+								nulls[4] = 'n';
+							else
+								nulls[4] = ' ';
```
steve-chavez, line 331: The [SPI_execute_with_args](https://www.postgresql.org/docs/current/spi-spi-execute-with-args.html) function uses this interface for specifying nulls - `'n'` is a NULL and `' '`(whitespace char) is a non-null. Weird, but it is what it is.

test/expected/test_http_get.out
```diff
@@ -24,3 +24,10 @@ rollback;
     -- from
         -- request;
 -- rollback;
+-- TODO run these tests:
+-- request with invalid protocol
+-- select net.http_get('net://supabase.io');
```
steve-chavez, line 29: Adding the test runs as a TODO for now.

sql/pg_net--0.1.sql
olirice: I think `error_msg` wants to be on `net.http_response` rather than the request queue. 

Currently, all errors, including local ones like `timed_out`, are persisted in `net.http_response`. If we add an `error_msg`to the request queue we'll have to check in the `http_request_queue` table in the `collect_response` function any time there is no value in `http_response`. We also won't be able to expose the errors that we find because the return type is `http_response` which has no field for `error_msg`

In other words, there is no public API to expose error messages in the request queue

test/sql/test_http_get.sql
```diff
@@ -22,3 +22,11 @@ rollback;
     -- from
         -- request;
 -- rollback;
+
+-- TODO run these tests:
```
olirice, line 26: We'll need a `select net.collect_response` under each of these to be confident that the error messages that get output are what we expect

olirice (changes requested):

test/sql/test_http_get.sql
steve-chavez: > I think error_msg wants to be on net.http_response rather than the request queue.
> In other words, there is no public API to expose error messages in the request queue

Yeah, you're right. I'll change the error message to be on the response table.

sql/pg_net--0.1.sql
steve-chavez: Yeah, you're right. I'll change the error message to be on the response table.

test/sql/test_http_get.sql
steve-chavez: Hm. You mean I need to change something here? I've just left these as comments for now.

sql/pg_net--0.1.sql
steve-chavez: > I think error_msg wants to be on net.http_response rather than the request queue.

@olirice On second thought, I don't think so. This is a **request error**, if I were to insert this on the response table that would mean it has to have a `status_code` and actually it doesn't(it's NULL) when there's a request error. 

There's probably more stuff that has to be changed from the `http_request_queue`. For starters the `timeout_milliseconds` should be a global config setting, not a column, the same goes for `delete_after`. Removing the `queue` from the name would also be good I think(queue is the use we give to the table, but otherwise is a regular entity).

Anyway, that's a lot of changes. I think they should be done on another PR. As it is, this PR is useful because it makes the extension more robust.

steve-chavez: `timed_out bool` could also be on the `http_request_queue` as it's also a property of the request - there is **no response** yet.

steve-chavez: >  If we add an error_msgto the request queue we'll have to check in the http_request_queue table in the collect_response

Hm, maybe the `collect_response` interface can change as well. If you notice, right now the worker does a LEFT JOIN, ideally it'd just do a SELECT over the request queue, less work inside would be better for performance/robustness.

test/sql/test_http_get.sql
olirice: They're commented out so no need to change them for now, only mentioning that we'll need to adjust these tests once there's a way to solve the multi-transaction problem

sql/pg_net--0.1.sql
steve-chavez: @olirice Right now I do this query to avoid repeating invalid requests:

```sql
SELECT
q.id, q.url, q.error_msg
FROM net.http_request_queue q\
LEFT JOIN net.http_response r ON q.id = r.id
WHERE r.id IS NULL AND q.error_msg is NULL
-- when the response is null and the request has no error message(valid)
```

How would I avoid processing invalid requests when changing the `error_msg` to response.table?

> I think error_msg wants to be on net.http_response rather than the request queue.

Actually it seems to me that `error_msg` wants to be on the request queue - since it's a **request error** and not a response(has no status code nor headers nor body).

olirice: > How would I avoid processing invalid requests when changing the error_msg to response.table?

If `error_message` was on the response, we'd then have a response record on failure so the query would be
```sql
select
   q.id, q.url
from
   net.http_request_queue q
   left join net.http_response r
      on q.id = r.id
where
    r.id is null
```

> Actually it seems to me that error_msg wants to be on the request queue - since it's a request error and not a response(has no status code nor headers nor body)

I see where you're coming from. Since request to response has a 1:1 relationship, and both tables are internal, I'm not too concerned about what the structure in the DB looks like. What's important to figure out is how to expose client/request errors & server/http side errors to users coherently through the public API

The public method for retrieving a response is `net.collect_response`, which has the return type `net.http_resposne`. Since that record would not include client/request errors, we don't have a way to let the user know when a client/request error has occurred.

In a perfect world we'd wrap the http response in [result type](https://doc.rust-lang.org/std/result/) and that's what would be returned from `net.collect_response`

```sql
-----------
-- TYPES --
-----------

-- A response from an HTTP server (abbreviated)
create type http_response as (request_id bigint, http_status_code int, contents text);

-- Potential states of an HTTP Request
create type request_status as enum ('PENDING', 'SUCCESS', 'ERROR');

-- State wrapper around (any protocol) responses
create type net_result as (status request_status, response http_response, msg text);

--------------
-- Examples --
--------------

-- A (non-http) error has occured e.g. timeout
select ('ERROR', null, 'client side timeout occurred')::net_result;
select ('ERROR', null, 'invalid protocol')::net_result;

-- Has not been processed
select ('PENDING', null, 'request has not been processed')::net_result;

-- Successful request resulting in http 404
select ('SUCCESS', (1, 404, 'Not Found')::http_response, 'ok')::net_result;

-- Successful request resulting in http 200
select ('SUCCESS', (1, 200, 'Some data')::http_response, 'ok')::net_result;
```

So in those examples, `net.collect_response` would return a `net_result` (`net.result`). I avoided this previously because ORMs don't always have good support for enums and composite types. Would postgrest be able to handle that response type ok @steve-chavez?

If so, that wrapping approach feels like it will be the most flexible for us going forward. What do you think?

steve-chavez: > If error_message was on the response, we'd then have a response record on failure so the query would be

Cool. I've changed the error message to be on the response table.

>  I avoided this previously because ORMs don't always have good support for enums and composite types. Would postgrest be able to handle that response type ok @steve-chavez?

Yeah, it should be fine. PostgREST doesn't use an ORM, if the types can be represented as json it should be fine. Though I don't know if this extension or its methods will be used through PostgREST, they seem more internal to me.

steve-chavez:
@olirice Is this good to merge?

PR #16, opened 2021-07-26
Isolation tests

olirice:
Use python/pytest for isolation tests since pg_isolation_regress isn't available and multiple transactions are needed

### DIFF
--- a/src/worker.c
+++ b/src/worker.c
@@ -187,7 +187,7 @@ worker_main(Datum main_arg)
 		/* Wait 10 seconds */
 		WaitLatch(&MyProc->procLatch,
 					WL_LATCH_SET | WL_TIMEOUT | WL_EXIT_ON_PM_DEATH,
-					10000L,
+					200L,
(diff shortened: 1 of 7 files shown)


src/worker.c
```diff
@@ -187,7 +187,7 @@ worker_main(Datum main_arg)
 		/* Wait 10 seconds */
 		WaitLatch(&MyProc->procLatch,
 					WL_LATCH_SET | WL_TIMEOUT | WL_EXIT_ON_PM_DEATH,
-					10000L,
+					200L,
```
olirice, line 190: @steve-chavez this is the only non-test related change

steve-chavez: Do you think the worker should be called this often? Perhaps a 1 second interval would be better.
```suggestion
					1000L,
```

steve-chavez: Otherwise it will execute many `SELECT .. LEFT JOIN` queries too often.

Issue #5, opened 2021-07-23
Look into body response as bytea


kiwicopple:
# Chore

## Describe the chore

Currently it's text

PR #17, opened 2021-07-27
feat: support POST

### LINKED ISSUE #6: Add support for http POST

soedirgo:

Feature

## What is the new behavior?

Support POST requests via `net.http_post()`. Right now the function is mostly copypasting from `http_get()`, but I think in the future this should be refactored to use a (public?) generic function `net.http()`.

@olirice the order of arguments to `net.http_post()` is a bit out of place since required arguments have to be listed first - let me know if you want them changed. Also I see I'm failing the pytest stuff - anything I should look at? Do you want me to add a test as well? (haven't looked at how it's setup)

@steve-chavez CMIIW, we currently ignore `params` and `headers` in the request right? Atm `content_type` is handled separately, but I think we should handle it together with `headers`. In retrospect, I probably should've omitted `content_type` in the SQL too... (was following pgsql-http) Also, let me know if I've freed the palloc'd stuff properly.

## Additional context

Closes #6.

## TODOs

- [x] merge `Content-Type` into `headers`
- [x] convert `body` to byte array to align with #5
- [x] `http_post` fails when `body` is null
- [ ] check memory leak

### DIFF
--- a/sql/pg_net--0.1.sql
+++ b/sql/pg_net--0.1.sql
@@ -15,7 +15,7 @@ create table net.http_request_queue(
     params jsonb not null,
     headers jsonb not null,
     content_type text,
-    body text,
+    body bytea,
     timeout_milliseconds int not null,
     -- Available for delete after this date
     delete_after timestamp not null
@@ -108,7 +108,7 @@ create or replace function net.http_post(
     -- ContentType of the POST request
     content_type text,
     -- body of the POST request
-    body text,
+    body bytea,
     -- key/value pairs to be url encoded and appended to the `url`
     params jsonb DEFAULT '{}'::jsonb,
     -- key/values to be included in request headers
--- a/sql/pg_net--0.1.sql
+++ b/sql/pg_net--0.1.sql
@@ -14,7 +14,6 @@ create table net.http_request_queue(
     url text not null,
     params jsonb not null,
     headers jsonb not null,
-    content_type text,
     body bytea,
     timeout_milliseconds int not null,
     -- Available for delete after this date
@@ -91,8 +90,8 @@ declare
     request_id bigint;
 begin
     -- Add to the request queue
-    insert into net.http_request_queue(method, url, params, headers, content_type, body, timeout_milliseconds, delete_after)
-    values ('GET', url, params, headers, null, null, timeout_milliseconds, timezone('utc', now()) + ttl)
+    insert into net.http_request_queue(method, url, params, headers, timeout_milliseconds, delete_after)
+    values ('GET', url, params, headers, timeout_milliseconds, timezone('utc', now()) + ttl)
     returning id
     into request_id;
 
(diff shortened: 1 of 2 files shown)


sql/pg_net--0.1.sql
```diff
@@ -101,6 +110,53 @@ begin
 end
 $$;
 
+-- Interface to make an async request
+-- API: Public
+create or replace function net.http_post(
+    -- url for the request
+    url text,
+    -- ContentType of the POST request
+    content_type text,
```
olirice, line 119: Since we have the `headers` parameter this can be removed. It'd be a 
```json
...
headers := '{"Content-Type": "text/html; charset=UTF-8"}'
...
```

olirice:
@soedirgo the background worker crashes while running test suite
Screen Shot 2021-07-27 at 3 52 21 AM

olirice (changes requested):

steve-chavez:
> CMIIW, we currently ignore params and headers in the request right? Atm content_type is handled separately, but I think we should handle it together with headers

@soedirgo Yeah, both ignored. Also agree with the `content-type` change, Oli has opened an issue for tracking that https://github.com/supabase/pg_net/issues/23.

soedirgo:
Both POST and headers should be supported now. As for the remaining issues, they can be handled here or in a separate PR:
- POST with body null fails despite null checking. Something similar happens when using null body for GET (which is legal HTTP, to my dismay)
- memory leak (labeled FIXME), I also suspect we may have other leaks given how many pfrees there are in pgsql-http (even for datum deserializations): https://github.com/pramsey/pgsql-http/blob/1e430f5869234e0c60625eb15a4f7c166cae1abe/http.c#L1006-L1009

steve-chavez:
> non-issue, this is because the functions are defined with strict

Functions with `strict` are bad for performance, as noted in this [postgrest issue](https://github.com/PostgREST/postgrest/issues/1294). I think that should be removed.

> memory leak (labeled FIXME), I also suspect we may have other leaks given how many pfrees there are in pgsql-http (even for datum deserializations)

Yeah, true. Ideally we'd integrate valgrind and also run it on CI.

src/worker.c
```diff
@@ -102,7 +103,30 @@ header_cb(void *contents, size_t size, size_t nmemb, void *userp)
...
+	char *header;
+
+	iterator = array_create_iterator(array, 0, NULL);
+
+	while (array_iterate(iterator, &value, &isnull))
+	{
+		if (isnull) continue;
+
+		header = TextDatumGetCString(value);
+		elog(DEBUG1, "Request header \"%s\"\n", header);
```
darora, line 121: probably don't want to inadvertently log auth info

soedirgo: Makes sense. With the default log level (`WARNING`) this wouldn't be printed but this is redundant anyway when setting `CURLOPT_VERBOSE`.

PR #27, opened 2021-07-28
Change response body from text to bytea

steve-chavez:
Fixes https://github.com/supabase/pg_net/issues/5.

I've checked the result by downloading an image:

```sql
select net.http_get('https://supabase.io/new/images/logo-dark.png');
\copy (select encode(body, 'hex') from net._http_response where id = 1) to 'supa.hex';
```
```bash
xxd -p -r supa.hex > supa.png
## open supa.png with any image viewer
```
(this way of getting the bytea to a file is mentioned in https://stackoverflow.com/a/6731452/4692662)

With this change, plain text media types have to be encoded. This can be done in pure SQL:

```sql
select net.http_get('https://supabase.io');
select encode(body, 'escape') from net._http_response where id = 1;
-- <!DOCTYPE html><html lang="en"....
```

We'd need to accommodate the SQL interface, perhaps it should handle `text/*` and some `application/*` media types for knowing when to encode.

### DIFF
--- a/src/worker.c
+++ b/src/worker.c
@@ -56,7 +56,7 @@ body_cb(void *contents, size_t size, size_t nmemb, void *userp)
 {
 	size_t realsize = size * nmemb;
 	StringInfo si = (StringInfo)userp;
-	appendBinaryStringInfo(si, (const char*)contents, (int)realsize);
+	appendBinaryStringInfoNT(si, (const char*)contents, (int)realsize);
 	return realsize;
 }
 
@@ -353,6 +353,8 @@ worker_main(Datum main_arg)
 											false, 1) != SPI_OK_INSERT)
 							{
 								elog(ERROR, "SPI_exec failed: %s", query_insert_response_bad.data);
+
+
 							}
 						} else {
 							int argCount = 6;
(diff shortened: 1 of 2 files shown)


src/worker.c
```diff
@@ -379,12 +382,14 @@ worker_main(Datum main_arg)
...
-							argValues[2] = CStringGetDatum(cdata->body->data);
-							if(cdata->body->data[0] == '\0')
-								nulls[2] = 'n';
-							else
-								nulls[2] = ' ';
+							argTypes[2] = BYTEAOID;
+
+							body_data = (bytea *) palloc(cdata->body->len + VARHDRSZ);
+							SET_VARSIZE(body_data, cdata->body->len + VARHDRSZ);
+							memcpy(VARDATA(body_data), cdata->body->data, cdata->body->len);
```
steve-chavez, line 389: This way of converting a `char*` to `bytea` is done in many parts of the postgres code([example](https://github.com/postgres/postgres/blob/8709228775b549a2388e6568f463a4115e5a4c10/contrib/pageinspect/heapfuncs.c#L389-L391)).

steve-chavez:
We'll hold back merging this one for this release. As @olirice mentioned internally, it's clunky for the client to escape the body since it needs to handle encoding:

```sql
select
    encode((x.response).body, 'escape')::jsonb body_json
from
    net.http_collect_response(1, async:=false) x;
```

A new interface like the following needs to be made:

```sql
select net.http_collect_response_text(request_id:=1)
select net.http_collect_response_bin(request_id:=1)
select net.http_collect_response_json(request_id:=1)
```

But since the most common use case will be getting plain text responses, this can wait.

PR #30, opened 2021-07-29
feat: support params directly in url

soedirgo:
https://curl.se/libcurl/c/libcurl-url.html

also change append query code with curl.

leaving `_urlencode` in to keep the tests passing

also granted everything to (hardcoded) `postgres` - probably not correct but we'll cross that bridge when we come to it


soedirgo:
Merging this now so I can test it on local - feel free to revert if there's an issue

PR #32, opened 2021-07-29
http_post json only

olirice:
text only requests for now


soedirgo:
Since the function is polymorphic on `body` type, maybe we can leave out the `Content-Type` check? So we set it automatically to `application/json`, putting it at the end of `headers` so it overwrites what the user has set.

PR #34, opened 2021-07-29
Adds TTL in worker

steve-chavez:
TTL is `3 days` by default, but can be configured with:

```sql
alter system set pg_net.ttl TO '5 minutes';
select pg_reload_conf();

show pg_net.ttl;
-- 5 minutes
```

### DIFF
--- a/test/fixtures.sql
+++ b/test/fixtures.sql
@@ -1,1 +1,3 @@
 create extension pg_net;
+alter system set pg_net.ttl TO '2 seconds';
+select pg_reload_conf();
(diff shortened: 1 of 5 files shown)


test/test_http_requests_deleted_after_ttl.py
```diff
@@ -3,45 +3,45 @@
 import pytest
 from sqlalchemy import text
 
-
-@pytest.mark.skip(reason="pending implementation")
 def test_http_requests_deleted_after_ttl(sess):
```
steve-chavez, line 6: This test now passes. Proves it's working.

test/fixtures.sql
```diff
@@ -1 +1,3 @@
 create extension pg_net;
+alter system set pg_net.ttl TO '2 seconds';
```
steve-chavez, line 2: 2 seconds TTL for all the tests.

olirice (approved):

PR #38, opened 2021-08-03
fix: allow usage of double quotes within json bodies

darora:
A valid json object that includes a double quote escaped within a string causes an error.

Not sure if this is the best way to fix this!

```
postgres=# select json_build_object('foo', 'b"ar')::text;
 json_build_object
-------------------
 {"foo" : "b\"ar"}
(1 row)

postgres=# select json_build_object('foo', 'b"ar')::text::bytea;
ERROR:  invalid input syntax for type bytea

postgres=# select convert_to(json_build_object('foo', 'b"ar')::text, 'UTF8');
              convert_to
--------------------------------------
 \x7b22666f6f22203a2022625c226172227d
(1 row)
```

### DIFF
--- a/sql/pg_net--0.1.sql
+++ b/sql/pg_net--0.1.sql
@@ -192,7 +192,7 @@ begin
         'POST',
         net._encode_url_with_params_array(url, params_array),
         headers,
-        body::text::bytea,
+        convert_to(body::text, 'UTF8'),
         timeout_milliseconds
     )
     returning id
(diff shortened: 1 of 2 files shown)


soedirgo:
Thanks @darora! The problem is `text::bytea` is expecting `escape` format for textual repr of `bytea`. You can only escape either a backslash or an octet, so `\"` is invalid (and the fix is correct - should have used `convert_to`).

LGTM! Waiting for the others to chime in.

olirice (approved):

PR #41, opened 2021-08-30
Release 0.2

soedirgo:


steve-chavez:
@soedirgo I see a

```
Rename:
sql/pg_net--0.1.sql → sql/pg_net--0.2.sql
```

I think it'd be better to keep a single `pg_net.sql` without the version, this one would be the complete sql at any version(I've seen something similar done for [pgvector](https://github.com/ankane/pgvector/tree/master/sql)).

soedirgo:
Thanks! Updated the PR.

steve-chavez (approved):

Issue #26, opened 2021-07-28
Address memory leak


soedirgo:
https://github.com/supabase/pg_net/pull/17

- [ ] use valgrind
- [ ] set up in CI

steve-chavez:
Can confirm there's a leak when doing some load testing(27.5% memory consumed after doing many requests, and it doesn't go down).

![2021-10-15-175308_1005x481_scrot](https://user-images.githubusercontent.com/1829294/137563217-e33247ab-7bf3-4b38-b618-0e7ebbc9ebd7.png)

steve-chavez:
I've managed to run valgrind on pg_net.

### Observations

- Gives a lot of false positives, even with the suppressions mentioned in [pg wiki](https://wiki.postgresql.org/wiki/Valgrind). Function calls for registering the background worker will show as leaks. 

```
 584 bytes in 1 blocks are possibly lost in loss record 42 of 70
    at 0x4838753: malloc (in /nix/store/llwiff4i5dwchnzlshdssdsnb0d04pbh-valgrind-3.16.1/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
    by 0x75235F: RegisterBackgroundWorker (in /nix/store/kx00x4h14m8zv0vzf7dzpjfm4n0wr006-postgresql-and-plugins-13.1/bin/.postgres-wrapped)
    by 0x773FFB: ApplyLauncherRegister (in /nix/store/kx00x4h14m8zv0vzf7dzpjfm4n0wr006-postgresql-and-plugins-13.1/bin/.postgres-wrapped)
    by 0x7618AD: PostmasterMain (in /nix/store/kx00x4h14m8zv0vzf7dzpjfm4n0wr006-postgresql-and-plugins-13.1/bin/.postgres-wrapped)
    by 0x4FF512: main (in /nix/store/kx00x4h14m8zv0vzf7dzpjfm4n0wr006-postgresql-and-plugins-13.1/bin/.postgres-wrapped)
```
- Valgrind will only help for regular `malloc` calls not postgres `palloc` calls, as mentioned in this [blog post](https://blog.cleverelephant.ca/2008/08/valgrinding-postgis.html)(note that we cannot run postgres in single user mode as shown here, because we need background workers)
- It does show some curl leaks, which happen on some parts we already suspected.

```
156 (64 direct, 92 indirect) bytes in 4 blocks are definitely lost in loss record 50 of 176
[...]

PR #50, opened 2021-10-21
feat: libuv

### LINKED ISSUE #13: Insert responses as soon as available
Currently, if 100 requests are made in a single transaction (or roughly at the same time) when the worker polls for new work it sends all 100 to the multi curl worker at the same time and doesn't write any of the responses to the `net.http_response` table until all 100 are complete. If any 1 of those servers is slow to respond, it may take up to 30 seconds (the current local timeout) before any responses are available.

Ideally, each response would be written to `net.http_response` as soon as its available

### LINKED ISSUE #46: Use `curl_multi_socket_action` with libevent instead of `curl_multi_perform`
Differences laid out here: https://curl.se/libcurl/c/libcurl-multi.html.

Could potentially help with https://github.com/supabase/pg_net/issues/13.

soedirgo:
Fixes #46, #13.

This PR integrates the [libuv](https://github.com/libuv/libuv) event loop with `curl_multi_socket_action` to replace the `curl_multi_perform`-based approach. Incremental changes wasn't really possible, so this is practically a half-rewrite (sorry about the mess!)

The main changes are:
- Use libuv as the event loop. Previously we also have a loop which mostly dealth with TTL and submitting new requests to curl. We integrate these logic into an [idle handle](https://docs.libuv.org/en/latest/guide/utilities.html#idler-pattern) - basically these let you run something once per loop iteration.
- Unlike the previous `curl_multi_perform` implementation, we now submit requests to curl one by one, each in its own transaction. This fixes #13 and sidesteps the need for the `batch_size` GUC, so I omitted it.
- The previous implementation stores per-request information in a global hash table. We sidestep the need for one by instead using a "request context" that is associated with each curl easy handle using `CURLOPT_PRIVATE`.
- Handles all memory leaks that I can eyeball (whatever that's worth).
- The request/response handling logic is mostly left unchanged.
- Adds a basic `.clang-format`. Feel free to configure it if you prefer.

Caveats:
- Fundamentally, event loops make understanding runtime behavior harder by exploding the state space of the program. Hopefully this is mitigated by judicious `elog`s, but it's not perfect.
- There's one more known failure mode, explained [here](https://github.com/supabase/pg_net/blob/8480aac5e453b85ee9358581add35d00aa1b2720/src/worker.c#L496-L505).

Blockers:
- replicate & fix the issue with invalid URLs & https

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,2 @@
+BasedOnStyle: LLVM
+IndentWidth: 4
--- a/Makefile
+++ b/Makefile
@@ -20,5 +20,8 @@ EXTRA_CLEAN = sql/$(EXTENSION)--$(EXTVERSION).sql
 PG_CONFIG = pg_config
 SHLIB_LINK = -lcurl -luv
 
+# Find <curl/curl.h> and <uv.h> from system headers
+PG_CPPFLAGS := $(CPPFLAGS)
+
 PGXS := $(shell $(PG_CONFIG) --pgxs)
 include $(PGXS)
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,7 @@ sql/$(EXTENSION)--$(EXTVERSION).sql: sql/$(EXTENSION).sql
 EXTRA_CLEAN = sql/$(EXTENSION)--$(EXTVERSION).sql
 
 PG_CONFIG = pg_config
-SHLIB_LINK = -lcurl
+SHLIB_LINK = -lcurl -luv
 
 PGXS := $(shell $(PG_CONFIG) --pgxs)
 include $(PGXS)
--- a/nix/pg_net.nix
+++ b/nix/pg_net.nix
@@ -1,9 +1,9 @@
-{ stdenv, postgresql, curl }:
+{ stdenv, postgresql, curl, libuv }:
 
 stdenv.mkDerivation {
   name = "pg_net";
 
-  buildInputs = [ postgresql curl ];
+  buildInputs = [ postgresql curl libuv ];
 
   src = ../.;
 
--- a/sql/pg_net--0.2--0.3.sql
+++ b/sql/pg_net--0.2--0.3.sql
@@ -1,0 +1,63 @@
+truncate net.http_request_queue;
+truncate net._http_response;
+
+drop index created_idx;
+alter table net.http_request_queue drop created;
+
+alter table net._http_response drop constraint _http_response_id_fkey;
+alter table net._http_response add created timestamptz not null default now();
+create index on net._http_response (created);
+
+create or replace function net.http_collect_response(
+    -- request_id reference
+    request_id bigint,
+    -- when `true`, return immediately. when `false` wait for the request to complete before returning
+    async bool default true
+)
+    -- http response composite wrapped in a result type
+    returns net.http_response_result
+    strict
+    volatile
+    parallel safe
+    language plpgsql
+as $$
+declare
+    rec net._http_response;
+    req_exists boolean;
+begin
+
+    if not async then
+        perform net._await_response(request_id);
+    end if;
+
+    select *
+    into rec
+    from net._http_response
+    where id = request_id;
+
+    if rec is null then
+        -- The request is either still processing or the request_id provided does not exist
+
+        -- TODO: request in progress is indistinguishable from request that doesn't exist
+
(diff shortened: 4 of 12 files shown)


soedirgo:
Forgot to reply to the above 😅

> Certainly we don't need top performance, but I assume if it's too slow users would complain that the changes are never reaching their functions.

Right, my point was we could start by making fundamental design decisions that improve performance (`curl_multi_socket`, shared mem, etc.) and leave the grunt work of optimizing later, otherwise we'd end up in a local maximum. Right now users are complaining about the robustness aspect rather than it being slow.

> Really weird, any ideas to what might be happening there? I've noted the worker is crashing/restarting in those loops.

Don't quite remember what the issues are in last week's commits, but generally I find that once active requests rise into the thousands, weird stuff start to happen.

> But considering the load tests seems this is not the case with the libuv version too(it might even be less robust for now).

Yeah, debugging this version has been pretty nasty. The jury's still out though if we can sort out the major blockers with `curl_multi_perform` - I'll check out your branch once it's pushed 👍

> Btw, note that you don't need NixOS to use Nix

Yeah I actually used nix-darwin until a few weeks ago - I never got pg_net's nix-shell to work because of some issues with stdenv.

---

I also have further updates:
- I reworked the queue logic in worker to `DELETE` every row processed - this allows the table to stay small so scans don't end up spiking CPU%. This needed some change in the schema though - in particular, right now we don't know whether a particular request id is being processed vs. doesn't exist, so I had to disable tests dealing with `PENDING`. (On that note, on `master` using request id past the TTL will do the same thing, so we have to fix this anyhow.)
- I used `CURLMOPT_MAX_TOTAL_CONNECTIONS` rather than limiting rows retrieved from the queue. I find that this works better for keeping active connections high enough to be performant and low enough to not trigger the "weird stuff", while still solving #13. Inserting responses is still a bottleneck though.
- I haven't added support & disabled tests for timeout - didn't work out of the box, probably need to relax the timeout & TTL numbers.

steve-chavez:
> I was able to get to 100+rps with this version - would be great if you could try replicating this.

6K reqs in 67 seconds ~= 88 req/s on the `t3a.nano`. Seems good :+1:. 

Some observations:

- The CPU usage is now capped at around 3%, which is good as is less than the batch size version(20%).
- There appears to be a memory leak now. When I do 6K requests, I get a %MEM of 36.3% and this doesn't go down, I've tried truncating the `http._net_response` table just in case but no dice.
- When I try to do 20K reqs,  this causes an OOM kill reliably and the requests seem to be forgotten(worker doesn't retry them on startup and the net.http_request_queue is empty). This looks related to the memory leak above.

> I used CURLMOPT_MAX_TOTAL_CONNECTIONS rather than limiting rows retrieved from the queue. I find that this works better for keeping active connections high enough to be performant and low enough to not trigger the "weird stuff", while still solving Insert responses as soon as available #13

This is much better. Before I could reliably reproduce the weird `systemd-coredump` issue(https://github.com/supabase/pg_net/pull/50#issuecomment-954102750) but now during my load tests I only saw it a couple of times when doing 6K reqs, on the first load test and after a `systemctl restart postgresql`. But this is not deterministic at all, maybe it will be gone when the memory leak is solved.

soedirgo:
> There appears to be a memory leak now.

Yikes. Well, I'm at my wit's end - I suspect I'll need mtrace/valgrind/etc. to track this down.

> requests seem to be forgotten(worker doesn't retry them on startup and the net.http_request_queue is empty)

This is expected as we do `DELETE RETURNING` now. Also related to the point about retrying above.

I'm not sure where we are re: solving #13 & #26 - do you have a version with `curl_multi_perform` that solves these?

steve-chavez:
> I'm not sure where we are re: solving #13 & #26 - do you have a version with curl_multi_perform that solves these?

Not yet. For #26, I wanted to have CI fail so the leak doesn't get repeated. I'm aiming to do that on https://github.com/supabase/pg_net/pull/55. Seems we'll need valgrind on CI after all since we have a new leak here.

steve-chavez (approved):
The `systemd-coredump` error still worries me a but but I think this is definitely the way forward. We can fix that later, the memory leak too.

I'd also like to do the shared memory approach and callback interface on top, this way I'll understand better some of the new cool things you've done here Bobbie. 

Merge time :rocket:

PR #55, opened 2021-11-09
Valgrind on CI

steve-chavez:
WIP for integrating valgrind into CI.

- [x] valgrind suppresions
- [ ] grep curl occurrence
- [ ] make CI fail


steve-chavez:
If this is done, we can get inspiration from orioledb valgrind check: https://github.com/orioledb/orioledb/blob/main/ci/check_output.sh (does grep).

PR #56, opened 2021-11-12
Timeout

### LINKED ISSUE #36: `timeout_milliseconds` is ignored

steve-chavez:
Closes #36


steve-chavez:
Adding the timeout somehow slows down the requests, not sure why. Will debug that later, for now the shared memory/callback is more urgent as it could change the SQL interface.

I'll make another PR with just the refactor and nix changes.

Issue #66, opened 2022-09-06
worker crashes on many `Connection refused` errors


steve-chavez:
Only happens when there's a `connection refused` error like:

```bash
curl localhost:8888
curl: (7) Failed to connect to localhost port 8888 after 0 ms: Connection refused
```

And after some tries

```sql
select net.http_get('http://localhost:8888') from generate_series(1,10);

free(): invalid pointer
2022-09-05 20:17:45.207 -05 [180720] LOG:  background worker "pg_net worker" (PID 180727) was terminated by signal 6: Aborted
2022-09-05 20:17:45.207 -05 [180720] LOG:  terminating any other active server processes

# The error message varies among these
# munmap_chunk(): invalid pointer
# free(): double free detected in tcache 2
# free(): invalid pointer
```

Other types of error don't cause the worker to crash.

---

Happens somewhere in

https://github.com/supabase/pg_net/blob/824b16338fd82fb459af92a1baab8e6b292367e5/src/worker.c#L76-L104

Issue #59, opened 2022-01-14
some remarks


RekGRpth:
1) https://github.com/supabase/pg_net/blob/04dd22fb7cf267d55057deccdf70aed373dc7032/src/worker.c#L434-L436
No! idle_cb function is called in TopMemoryContext context, and it does not pfree palloc'ed like
https://github.com/supabase/pg_net/blob/04dd22fb7cf267d55057deccdf70aed373dc7032/src/worker.c#L452
https://github.com/supabase/pg_net/blob/04dd22fb7cf267d55057deccdf70aed373dc7032/src/worker.c#L486
https://github.com/supabase/pg_net/blob/04dd22fb7cf267d55057deccdf70aed373dc7032/src/worker.c#L490
etc

RekGRpth:
2) https://github.com/supabase/pg_net/blob/04dd22fb7cf267d55057deccdf70aed373dc7032/src/worker.c#L147
https://github.com/supabase/pg_net/blob/04dd22fb7cf267d55057deccdf70aed373dc7032/src/worker.c#L360
https://github.com/supabase/pg_net/blob/04dd22fb7cf267d55057deccdf70aed373dc7032/src/worker.c#L488
https://github.com/supabase/pg_net/blob/04dd22fb7cf267d55057deccdf70aed373dc7032/src/worker.c#L492
https://github.com/supabase/pg_net/blob/04dd22fb7cf267d55057deccdf70aed373dc7032/src/worker.c#L510
postgres has own palloc function, so You should not use malloc!

RekGRpth:
3) https://github.com/supabase/pg_net/blob/04dd22fb7cf267d55057deccdf70aed373dc7032/src/worker.c#L549
uv library is great! but postgres has own event loop and You should use it like in pg_task
https://github.com/RekGRpth/pg_task/blob/e6b882618e238f05e3a3260ebf43f392ffd8b2c9/work.c#L59-L67
https://github.com/RekGRpth/pg_task/blob/e6b882618e238f05e3a3260ebf43f392ffd8b2c9/work.c#L788

RekGRpth:
4) https://github.com/supabase/pg_net/blob/04dd22fb7cf267d55057deccdf70aed373dc7032/src/worker.c#L530
really?! You work under postgres superuser in postgres database?! not-superusers should have power rights to access net.* tables. And how access from another databases? dblink?

RekGRpth:
[...]

PR #68, opened 2022-09-08
revert to non-libuv worker, fix all pending bugs

steve-chavez:
Immediate benefit is that it fixes https://github.com/supabase/pg_net/issues/66, added a test for proving it.

Also planning to add [mitmproxy's pathod](https://2qwesgdhjuiytyrjhtgdbf.readthedocs.io/en/latest/pathod/intro.html#pathod) for testing against other http faillures like disconnects. 

Will refine other things along the way.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -17,5 +17,5 @@ jobs:
       - name: Run pytest tests
         run: nix-shell --run "net-with-pg-13 python -m pytest -vv"
 
-      - name: Check C files are formatted
-        run: nix-shell --run "net-check-format"
+      #- name: Check C files are formatted
+        #run: nix-shell --run "net-check-format"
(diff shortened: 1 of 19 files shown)


steve-chavez:
This one should close all the [pending bugs](https://github.com/supabase/pg_net/issues?q=is%3Aissue+is%3Aopen+label%3Abug).

For https://github.com/supabase/pg_net/issues/59, the only relevant remark left was the [hardcoded database name](https://github.com/supabase/pg_net/issues/59#issuecomment-1012734035), which is now fixed. The other remarks are no longer relevant since libuv is gone, malloc is no longer used, etc.

For https://github.com/supabase/pg_net/issues/42, I wasn't able to reproduce locally but I did saw a case in production where after a `Server returned nothing (no headers, no data)` response the worker stopped working. I've added a test with `pathod` that proves we keep working if that happens.

https://github.com/supabase/pg_net/issues/26, is no longer reproducible but I've corrected a FIXME in the code where `curl_slist_free_all` wasn't being used.

The other bugs should be clear with the commits.

soedirgo (approved):

Issue #80, opened 2023-01-24
`net._http_response` rows do not respect TTL


steve-chavez:
### Problem

When having a TTL of:

```sql
show pg_net.ttl;

6 hours
```

Doing new requests succeed, but their response rows get deleted quickly(after 1 second):

```sql
select net.http_get('https://webhook.site/449fddab-5642-4645-9a5e-6c71f7fe9a8e'); 
select * from net._http_response;
-- 0 rows
```

This is because the worker inserts rows using a `now()` default value:

https://github.com/supabase/pg_net/blob/639b0498fd1835a59ab763c2a84a1dfe4c3a21dd/src/worker.c#L339-L340
https://github.com/supabase/pg_net/blob/639b0498fd1835a59ab763c2a84a1dfe4c3a21dd/sql/pg_net.sql#L41

And this `now()` is not getting updated. Every new insert uses an "old" `now()` value.

### Solution

Use `SetCurrentStatementStartTimestamp()` before:

https://github.com/supabase/pg_net/blob/639b0498fd1835a59ab763c2a84a1dfe4c3a21dd/src/worker.c#L205-L206

To make `now()` use the current time.

Issue #78, opened 2023-01-19
pg_net prevents `DROP database foo` from succeeding


steve-chavez:
Only happens on Supabase prod instances which have pg 15, when doing:

```sql
CREATE DATABASE foo;
DROP DATABASE foo;
```

`DROP DATABASE foo` gets locked with the following message:

```
still waiting for backend with PID 2576 to accept ProcSignalBarrier
```

PID 2576 was pg_net in this case.

---

To reproduce locally, we need a dummy function handling `SIGUSR1`.

(Otherwise the problem is not reproducible locally as pg_net doesn't handle SIGUSR1 explicitly, on Supabase production other extensions in `shared_preload_libraries` are likely triggering this effect.)

```c
static void
handle_sigusr1(SIGNAL_ARGS)
{
  elog(WARNING, "foo");
}

void
worker_main(Datum main_arg)
{
// ...
  pqsignal(SIGUSR1, handle_sigusr1);
}
```

```
$ net-with-pg-15 psql

postgres=# create database foo;
CREATE DATABASE
postgres=# drop database foo;
2023-01-19 02:36:38.827 -05 [463963] LOG:  checkpoint starting: immediate force wait
2023-01-19 02:36:38.830 -05 [463963] LOG:  checkpoint complete: wrote 92 buffers (0.6%); 0 WAL file(s) added, 0 removed, 0 recycled; write=0.003 s, sync=0.001 s, total=0.003 s; sync files=0, l
ongest=0.000 s, average=0.000 s; distance=4714 kB, estimate=4714 kB
2023-01-19 02:36:38.830 -05 [463968] WARNING:  foo

2023-01-19 02:36:43.835 -05 [463973] LOG:  still waiting for backend with PID 463968 to accept ProcSignalBarrier
2023-01-19 02:36:43.835 -05 [463973] STATEMENT:  drop database foo;
2023-01-19 02:36:48.840 -05 [463973] LOG:  still waiting for backend with PID 463968 to accept ProcSignalBarrier
2023-01-19 02:36:48.840 -05 [463973] STATEMENT:  drop database foo;

```
---

To solve this we'd need to handle ProcSignalBarrier, which is already done by `procsignal_sigusr1_handler` [here](https://github.com/postgres/postgres/blob/980ae173108e353045e5ab4a842bb21e9dfe6715/src/backend/storage/ipc/procsignal.c#L655-L656). We just need to add:

```c
pqsignal(SIGUSR1, procsignal_sigusr1_handler);
```

Issue #63, opened 2022-04-11
Allow http DELETE


psteinroe:
# Feature request

## Is your feature request related to a problem? Please describe.

We want to send a delete request to a serverless function within a Nextjs api to trigger the deletion of the entity at an external provider after it was deleted in our database.

## Describe the solution you'd like

Support http delete in addition to post and get.

## Describe alternatives you've considered

Adding another endpoint, e.g. /delete, and sending a post request there

tvogel:
See also #77 . Lack of DELETE makes it impossible to use pg_net to clean up supabase storage objects in UPDATE/DELETE Postgres triggers. Or am I missing something?

steve-chavez:
Can be used on v0.7 as:

```sql
        select net.http_delete(
            url:='https://httpbin.org/delete'
        ,   params:= '{"param-foo": "bar"}'
        ,   headers:= '{"X-Baz": "foo"}'
        );
```

This will be deployed to Supabase soon.

GaryAustin1:
Is `body` also supported in delete?  I have a delete storage task that uses bulk delete with the http extension right now.  It uses body for pathnames.  I don't really need to wait for a response in what I'm doing and pg_net would be ideal.

steve-chavez:
@GaryAustin1 Not supported, will track it on https://github.com/supabase/pg_net/issues/77

PR #81, opened 2023-01-24
To 0.7

steve-chavez:
- Fixes https://github.com/supabase/pg_net/issues/78
- Fixes https://github.com/supabase/pg_net/issues/80
- Closes https://github.com/supabase/pg_net/issues/63


pcnc (approved):

sql/pg_net.sql
```diff
@@ -232,7 +276,7 @@ create type net.http_response_result as (
 
 
 -- Collect respones of an http request
--- API: Public
+-- API: Private
```
soedirgo, line 279: Should we also append `_` to the function name?

soedirgo (approved):

sql/pg_net.sql
steve-chavez: Right, done.

GaryAustin1:
https://github.com/supabase/supabase/issues/12008
Could be unrelated, but webhooks are no longer installing on new instances sometime in the past few days.

GaryAustin1:
image
Is the error coming from this...
 `"query": "BEGIN;\n\n  -- Create pg_net extension\n  CREATE EXTENSION IF NOT EXISTS pg_net SCHEMA`

PR #91, opened 2023-05-18
Updated API docs, removed dated comments

TheOtherBrian1:
## What kind of change does this PR introduce?

Docs Update

## What is the current behavior?

The [API documentation](https://github.com/supabase/pg_net/blob/master/docs/api.md) does not include net.http_delete. Also each request function signature still contains a warning against the usage of timeout parameters, even though the bug that necessitated the warning had been resolved. 

Please link any relevant issues here.

## What is the new behavior?

- I removed the timeout warning comments from the request functions. 
- I added the net.http_delete function to the API documentation

## Additional context

Add any other context or screenshots.

Issue #74, opened 2022-10-26
Timeout was reached error with incorrect status: SUCCESS


tonyxiao:
# Bug report

1. Run supabase/postgres docker
2. Run sql
```sql
CREATE EXTENSION IF NOT EXISTS pg_net;
select
    net.http_post(
        url:='https://httpbin.org/post',
        body:='{"hello": "world"}'::jsonb
    );

select
  *
from
  net.http_collect_response(1);
```

@see

<a href="https://www.loom.com/share/57be14ed3bb541a889dfa2b0e0246f31">
    <p>pg_net: Timeout was reached - Watch Video</p>
    <img style="max-width:300px;" src="https://cdn.loom.com/sessions/thumbnails/57be14ed3bb541a889dfa2b0e0246f31-with-play.gif">
  </a>

Notice how the status says `SUCCESS`, but upon inspecting the `net._http_response` table, it says timeout was reached. Additionally it appears that the request was never made according to the remote server.

steve-chavez:
Thanks for the report. Note that there's a timeout(by default 2 seconds) you can add to the function.

```sql
select
    net.http_post(
        url:='https://httpbin.org/post',
        body:='{"hello": "world"}'::jsonb,
        timeout_milliseconds := 5000
    );
```

The SUCCESS is indeed a mistake.

tonyxiao:
I see. What's the default timeout on supabase itself? Or is it simply faster on supabase and therefore within the 2000ms timeframe?

steve-chavez:
Default it's the same - 2 seconds. 

> Or is it simply faster on supabase and therefore within the 2000ms timeframe?

Yeah, latency might be lower due to network proximity.

steve-chavez:
`http_collect_response` was made a private function on https://github.com/supabase/pg_net/pull/81.

For now it's better to inspect `net._http_response` directly. We're going to improve this on https://github.com/supabase/pg_net/issues/62.

Issue #62, opened 2022-02-15
Change table queue to in-memory queue and add callbacks


steve-chavez:
### Reasons

- For all the requests to finish, an INSERT on the `_http_response` table must be done, this reduces throughput. There are cases where the client doesn't care about the response so it doesn't need to be persisted. For example, [function hooks](https://supabase.com/blog/2021/07/30/supabase-functions-updates#function-hooks-alpha) don't do anything with the response.
- As discussed in https://github.com/supabase/pg_net/pull/50#issuecomment-953796674, we don't have to ensure delivery and do retries since we're just an HTTP client, so we don't really need to persist the requests into `http_request_queue`.
- The `http_request_queue` can grow big with many requests and cause production issues([internal link](https://support.supabase.org/a/tickets/4806)).
- Users create FKs to our private tables https://github.com/supabase/pg_net/issues/44, making the bg worker crash when trying to do the TTL.

### Proposal

Drop the `_http_response` and `http_request_queue` tables and instead use an in-memory queue, plus add two callbacks:

```sql
create or replace function net.http_get(
   url text,
-- ..,
   success_cb text default '',
   error_cb text default ''
)
returns void -- no id is returned now
as $$
-- ...
$$ language sql;
```

Which can be used like:

```sql
select net.http_get(
  url := 'http://domain.com',
, success_cb := $$ insert into my_table values ($1, $2, $3) $$ -- $1=status, $2=headers, $3=body
, error_cb := $$ do $_$begin raise exception 'Failed request on %: "%"', $1, $2; end$_$; $$ -- $1=url, $2=error_message
);
```

### Pros

- The callbacks are optional, so we don't have to start a transaction for each request if there are none provided. Throughput should be greatly increased from this.
- The `error_cb` can also be an insert/update on a table, so the request can be retried if needed.

---

@olirice @soedirgo WDYT?

soedirgo:
Ohh, so that's what you meant by callbacks. Nice! Also like that this makes the impl a lot simpler and more akin to typical HTTP clients.

How would go about implementing the in-memory queue? IIRC when I looked into it passing info from the client process to the bgworker process was nontrivial.

olirice:
that looks great!

Is it correct that all http status codes (e.g. including 500) would still go through the `success_cb` and the `error_cb` applies to stuff like timeouts?

-------

```sql
returns void -- no id is returned now
```

I think it would still be good to return a `uuid` and pass that through to the callbacks i.e.

```sql
[...]

PR #93, opened 2023-05-18
Updated SQL: Timed out responses are no longer recorded as "SUCCESSFUL"

TheOtherBrian1:

Bug fix

Timed out responses are labeled as successful. 

Please link any relevant issues here: [issue 74](https://github.com/supabase/pg_net/issues/74)


## What is the new behavior?

Time out responses are now recognized as "ERRORS"

## Additional context

The unexpected test failures in my recent pull requests ([1](https://github.com/supabase/pg_net/actions/runs/4987953205/attempts/5), [2](https://github.com/supabase/pg_net/pull/91)), which only modified documentation, prompted me to investigate the error. The root cause appears to be timeouts during test endpoint calls to https://httpbin.org. If the response exceeds the default 2-second timeout, `net._http_collect_response` returns a misleading success status:

```markdown
| status  | message | response |
| ------- | ------- | -------- |
| SUCCESS | ok      | (,,)     |
```
When the tests encounter a SUCCESS status, they anticipate the response field to be populated with the tuple `(<status_code>, <headers>, <body>)`. However, because timed-out responses are mislabeled as successful, this field remains unpopulated, leading to discrepancies. One fix would be to adjust the [worker.c](https://github.com/supabase/pg_net/blob/master/src/worker.c) to yield an 'ERROR' code upon timeouts, but ongoing discussions around a potential breaking update in [issue 74](https://github.com/supabase/pg_net/issues/74) and [issue 62](https://github.com/supabase/pg_net/issues/62) suggest caution.

While looking for an alternative solution, I ran two GET requests using the [Postman Echo API](https://learning.postman.com/docs/developer/echo-api/):

```sql
-- REQUEST_ID 81: TIMED OUT
SELECT net.http_get (
            'https://postman-echo.com/get?foo1=bar1&foo2=bar2',
            timeout_milliseconds := 1
) AS request_id;

-- REQUEST_ID 82: SUCCESSFUL
SELECT net.http_get (
            'https://postman-echo.com/get?foo1=bar1&foo2=bar2'
) AS request_id;
```
Here are the requests' records in the net._http_response table:

| id | status_code | content_type                    | headers         | content                                                                                                                                                                                                                                                                                                                                                                                                  | timed_out | error_msg           | created                       |
|----|-------------|---------------------------------|-----------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|-----------|---------------------|-------------------------------|
| 81 |             |                                 |                 |                                                                                                                                                                                                                                                                                                                                                                                                          |           | Timeout was reached | 2023-05-18 04:01:07.054579+00 |
| 82 | 200         | application/json; charset=utf-8 | [object Object] | `{ "args": { "foo1": "bar1", "foo2": "bar2" }, "headers": { "x-forwarded-proto": "https", "x-forwarded-port": "443", "host": "postman-echo.com", "x-amzn-trace-id": "Root=1-6465a4be-32a2128c278ce40d3879ff2f", "accept": "*/*", "api-key-header": "<API KEY>", "user-agent": "pg_net/0.7.1" }, "url": "https://postman-echo.com/get?foo1=bar1&foo2=bar2" }` | false     |                     | 2023-05-18 04:08:30.839851+00 |


I discovered that [net._http_collect_response](https://github.com/supabase/pg_net/blob/f7ea986b8241a9adbe278d4868875650a3f7db36/sql/pg_net.sql#L280), while determining success a response's status, overlooks error messages and assumes any recorded response denotes a successful execution. My proposed solution inspects the error message field for each request, and if a message is found or the request is absent, the function will return an 'ERROR' status. I modified only two lines within the function (denoted by comments) which should resolve this issue.


```sql

-- Collect responses of an http request
-- API: Private
create or replace function net._http_collect_response(
    -- request_id reference
    request_id bigint,
    -- when `true`, return immediately. when `false` wait for the request to complete before returning
    async bool default true
)
    -- http response composite wrapped in a result type
    returns net.http_response_result
    strict
    volatile
    parallel safe
    language plpgsql
    security definer
as $$
declare
    rec net._http_response;
    req_exists boolean;
begin

    if not async then
        perform net._await_response(request_id);
    end if;

    select *
    into rec
    from net._http_response
    where id = request_id;
    
    --  If no response or an error exists, return ERROR
    -- BELOW LINE MODIFIED: ADDED OR CONDITION
    if rec is null OR rec.error_msg IS NOT NULL then 
        -- The request is either still processing or the request_id provided does not exist

        -- TODO: request in progress is indistinguishable from request that doesn't exist

        -- No request matching request_id found
        return (
            'ERROR',
            -- BELOW LINE MODIFIED: ADDED COALESCE
            COALESCE(rec.error_msg, 'request matching request_id not found'),
            null
        )::net.http_response_result;

    end if;


    -- Return a valid, populated http_response_result
    return (
        'SUCCESS',
        'ok',
        (
            rec.status_code,
            rec.headers,
            rec.content
        )::net.http_response
    )::net.http_response_result;
end;
$$;
```

Also, I would like to ask: is there any reason why this function is commented as "Private"? It seems like it shouldn't be.

### DIFF
--- a/sql/pg_net.sql
+++ b/sql/pg_net.sql
@@ -305,7 +305,7 @@ begin
     from net._http_response
     where id = request_id;
 
-    if rec is null then
+    if rec is null or rec.error_msg is not null then
         -- The request is either still processing or the request_id provided does not exist
 
         -- TODO: request in progress is indistinguishable from request that doesn't exist
@@ -313,7 +313,7 @@ begin
         -- No request matching request_id found
         return (
             'ERROR',
-            'request matching request_id not found',
+            coalesce(rec.error_msg, 'request matching request_id not found'),
             null
         )::net.http_response_result;
 


steve-chavez:
> Also, I would like to ask: is there any reason why this function is commented as "Private"? It seems like it shouldn't be.

`net._http_collect_response` uses an infinte loop(`_await_response`) to wait for a response.

I recall an user was using `net._http_collect_response` on triggers, which of course caused timeouts on slow http responses. 

So yeah, that function is only meant for development purposes. If we really want to expose sync requests, we need to come up with a C implementation.

steve-chavez (approved):
Awesome PR, well explained :100:.

Issue #94, opened 2023-05-19
Test Failures due to Timeout


TheOtherBrian1:
# Bug report

- [ X ] I confirm this is a bug with Supabase, not with my own application. 
- [ X ] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

I noticed that my tests unexpectedly failed despite only making changes to the documentation. The root cause appears to be timeout errors resulting from the [httpbin](https://httpbin.org/) endpoints being excessively slow (over 2 seconds for response). These timeout errors are not accounted for in the tests, causing them to fail even with functional code.

## To Reproduce

To replicate, simply run the tests. A failure typically occurs due to the unanticipated timeout error.

## Expected behavior

The tests should retry requests if there is a timeout error.

steve-chavez:
> The root cause appears to be timeout errors resulting from the [httpbin](https://httpbin.org/) endpoints being excessively slow (over 2 seconds for response).

@TheOtherBrian1 True, just noticed that as well. I think we should change those tests to use our Nginx server(one is started for the test suite) endpoints or Postman endpoints.

TheOtherBrian1:
@steve-chavez I could modify the tests to work with the Postman Echo API. It's designed specifically for this kind of testing, and can be easily swapped with httpbin. Looking through the tests, it would probably only take me half an hour to an hour to implement. Timeouts are still bound to happen occasionally, but it would be an improvement.

steve-chavez:
@TheOtherBrian1 Cool. Please go ahead.

TheOtherBrian1:
@steve-chavez I added my changes to a [pull request](https://github.com/supabase/pg_net/pull/96), but it appears the tests are taking an unusually long amount of time (50 minutes at this point). By default Github actions will timeout after 6 hours, which seems excessive in this context. 

The [changes](https://github.com/supabase/pg_net/pull/96/files) I made were fairly mild, so I'm hesitant to think that I accidentally created an infinite loop that is holding up the server. It's possible that because the actions are dependent on the test code, the modifications interfered with their normal function. Either or, it is probably wise to lower the timeout limit.

TheOtherBrian1:
[...]

PR #99, opened 2023-08-08
clarify how to use pg_net on a different database

soedirgo:


soedirgo:
why is this flaky :(

TheOtherBrian1:
[The tests do not work because the requests time out before a response can be returned.](https://github.com/supabase/pg_net/issues/94) When the tests compare the expected return value to the message of a timeout error, it fails. The default max time for any request is 2 seconds. The endpoint, httpbin, is too slow. 

I tried changing the API endpoint from httpbin to postman's testing API in a [pull request](https://github.com/supabase/pg_net/pull/96) some time ago. I was hoping utilizing a more reliable endpoint would solve the issue. However, after Github actions took over an hour to execute the test, I closed my pull request. 

I'm not too familiar with Github actions, but perhaps modifying the testing code it relied on resulted in strange behaviors. Either or, if you can limit the max runtime for the actions, I think you should.

steve-chavez (approved):

PR #102, opened 2023-08-17
test: httpbin.org -> localhost:8080

steve-chavez:
Move all the tests from httpbin.org to local openresty server

Following up from https://github.com/supabase/pg_net/pull/99#issuecomment-1669975876

### DIFF
--- a/nix/nginx/conf/custom.conf
+++ b/nix/nginx/conf/custom.conf
@@ -11,3 +11,29 @@ location /really-slow-reply {
 location /echo-method {
   echo $request_method;
 }
+
+location /anything {
+  echo $is_args$query_string;
+}
+
+location /headers {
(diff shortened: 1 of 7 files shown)


nix/nginx/conf/custom.conf
```diff
@@ -11,3 +11,29 @@ location /really-slow-reply {
...
+  echo_read_request_body;
+  echo $request_body;
+}
+
+location /delete {
+  if ($request_method != 'DELETE'){
+      return 405;
+  }
+  echo_duplicate 1 $echo_client_request_headers$is_args$query_string;
+}
```
steve-chavez, line 39: This config is all there's to it. The tests are mostly kept as is.

soedirgo (approved):
🙌 Nice work!

PR #105, opened 2023-09-15
fix: check_worker_is_up

steve-chavez:
- [fix: check_worker_is_up](https://github.com/supabase/pg_net/commit/b6f6517a0d45ce98de2e610049a5da1f787aeda2)
  
  * Test it works, previously it only tested failure.

### DIFF
--- a/sql/pg_net.sql
+++ b/sql/pg_net.sql
@@ -20,7 +20,7 @@ create unlogged table net.http_request_queue(
 
 create or replace function net.check_worker_is_up() returns void as $$
 begin
-  if not exists (select pid from pg_stat_activity where backend_type = 'pg_net worker') then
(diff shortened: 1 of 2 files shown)


soedirgo:
I'm a bit wary about changing a function's signature. Can we instead create a new function (`net.is_worker_up()`) and keep the old function as is?

soedirgo:
We can also print a warning when the old function is run.

steve-chavez:
Ok, signature is kept as is.

I thought that `check_worker_is_up` would be confusing if it doesn't return anything. I just added a comment to clarify this.

steve-chavez:
I'll hold bumping to 0.7.3 after https://github.com/supabase/pg_net/pull/108 is merged

soedirgo (approved):

PR #109, opened 2023-09-19
bump to 0.7.3

steve-chavez:


steve-chavez:
@soedirgo After this is merged, I just need to push a tag and the release will work? Or are there other steps needed?

steve-chavez:
I think it should work, I see pg 16 on apt repos https://apt.postgresql.org/pub/repos/apt/pool/16/p/

soedirgo:
I *think* you only need to push a tag yeah

soedirgo (approved):

PR #114, opened 2023-11-02
fix: missing secdef in 0.5.1-0.6.sql

soedirgo:


steve-chavez (approved):

steve-chavez:
@soedirgo LGTM. What was the problem this caused though?

soedirgo:
There was a project that was on 0.2, and running `alter extension update` didn't make it reflect the latest version's schema.

PR #117, opened 2023-11-16
feat: allow reconfiguring pg_net w/o a db restart

steve-chavez:
Otherwise changing a pg_net config requires a restart of the whole database cluster.

Now users can do:

```sql
alter system set pg_net.ttl to '1 hour';
alter system set pg_net.batch_size to 500;

select net.worker_restart();
```

### DIFF
--- a/sql/pg_net.sql
+++ b/sql/pg_net.sql
@@ -353,5 +353,14 @@ begin
 end;
 $$;
 
+create or replace function net.worker_restart() returns bool as $$
+  select pg_reload_conf();
+  select pg_terminate_backend(pid)
+  from pg_stat_activity
+  where backend_type ilike '%pg_net%';
+$$
+security definer
+language sql;
+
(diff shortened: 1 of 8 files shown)


sql/pg_net.sql
```diff
@@ -353,5 +353,14 @@ begin
...
 
+create or replace function net.worker_restart() returns bool as $$
+  select pg_reload_conf();
+  select pg_terminate_backend(pid)
+  from pg_stat_activity
+  where backend_type ilike '%pg_net%';
+$$
+security definer
+language sql;
+
```
soedirgo, line 364: We also need to bump the version and add a migration

steve-chavez: Yes, but that can be done on another PR that bumps the version. I plan on adding a few more fixes before that.

soedirgo (approved):

PR #120, opened 2024-01-10
fix: permission denied for net tables 

### LINKED ISSUE #72: permission denied for table http_request_queue as an authenticated user
# Bug report

## Describe the bug

After pg_net got upgraded, it throws a permission denied for table http_request_queue when trying to make a request as an authenticated user.

## To Reproduce

Upgrade to the latest version and attempt to make a http request using the authenticated role.

## Expected behavior

Assumption is that any supabase extension works out of the box with the role and auth setup in the database. If it is expected behaviour, it should be documented somewhere.

steve-chavez:
Closes #72

Also bumps version to 0.8.0. Includes the change from https://github.com/supabase/pg_net/pull/123

### DIFF
--- a/sql/pg_net--0.7.3--0.8.0.sql
+++ b/sql/pg_net--0.7.3--0.8.0.sql
@@ -1,0 +1,142 @@
+create or replace function net.worker_restart() returns bool as $$
+  select pg_reload_conf();
+  select pg_terminate_backend(pid)
+  from pg_stat_activity
+  where backend_type ilike '%pg_net%';
+$$
+security definer -- needs SDF because of pg_terminate_backend
+language sql;
+
+grant usage on schema net to PUBLIC;
+grant all on all tables in schema net to PUBLIC;
+
+alter function net.http_get ( text, jsonb, jsonb, integer) security invoker;
+alter function net.http_post (text, jsonb, jsonb, jsonb, integer) security invoker;
+alter function net.http_delete (text, jsonb, jsonb, integer) security invoker;
+alter function net._http_collect_response ( bigint, boolean) security invoker;
+alter function net.http_collect_response ( bigint, boolean) security invoker;
+
+create or replace function net.http_get(
+    url text,
+    params jsonb default '{}'::jsonb,
+    headers jsonb default '{}'::jsonb,
+    timeout_milliseconds int default 5000
+)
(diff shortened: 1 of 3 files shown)


sql/pg_net--0.7.3--0.8.0.sql
```diff
@@ -0,0 +1,17 @@
...
+  select pg_reload_conf();
+  select pg_terminate_backend(pid)
+  from pg_stat_activity
+  where backend_type ilike '%pg_net%';
+$$
+security definer -- needs SDF because of pg_terminate_backend
+language sql;
+
+grant all on schema net to PUBLIC;
+grant all on all tables in schema net to PUBLIC;
```
soedirgo, line 11: This should be `usage` - we don't want people creating objects in the `net` schema

sql/pg_net--0.7.3--0.8.0.sql
```diff
@@ -0,0 +1,17 @@
+create or replace function net.worker_restart() returns bool as $$
+  select pg_reload_conf();
+  select pg_terminate_backend(pid)
+  from pg_stat_activity
+  where backend_type ilike '%pg_net%';
+$$
+security definer -- needs SDF because of pg_terminate_backend
+language sql;
+
+grant all on schema net to PUBLIC;
```
soedirgo, line 10: This should be `usage` - we don't want people creating objects in the `net` schema

steve-chavez: @soedirgo Fixed it. Do you see anything else that's missing?

soedirgo:
Looking at the linked issue, if some tool is resetting the privileges on these functions, I don't think moving them to the tables would help.

pg_cron is an example of the contrary, because it only grants `select` (and even then, only for `cron.job` and not for `cron.job_run_details`) and inserts are done through a kinda-sorta secdef function.

soedirgo:
If we're sure about granting to `public` (which I'm still not sure what the implications are) it seems a more incremental change would be to keep the functions as secdef but grant `execute` to `public`.

steve-chavez:
Right, maybe we should leave the privileges as is. #72 has been rare until now.

steve-chavez:
Follow up on https://github.com/supabase/pg_net/pull/158

PR #129, opened 2024-03-14
fix: usage of net schema when the postgres role misses privileges

steve-chavez:
Related to https://github.com/orgs/supabase/discussions/21450.

## Notes

- Only the usage privilege on schema net should be required.
- Also add GRANT on sequences to PUBLIC. See https://github.com/citusdata/pg_cron/blob/main/pg_cron--1.4--1.4-1.sql.

### DIFF
--- a/test/test_privileges.py
+++ b/test/test_privileges.py
@@ -1,0 +1,92 @@
+import pytest
+from sqlalchemy import text
+
+def test_net_postgres_usage(sess):
+    """Check that the postgres role can use the net schema by default"""
+
+    role = sess.execute(text("select current_user;")).fetchone()
+
+    assert role[0] == "postgres"
+
+    # Create a request
+    (request_id,) = sess.execute(text(
+        """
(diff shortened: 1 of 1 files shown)


test/test_privileges.py
```diff
@@ -0,0 +1,92 @@
...
+            revoke all on all tables in schema net from postgres;
+
+            -- still, assume postgres has grant usage on the schema net
+            grant usage on schema net to postgres;
+
+            select net.http_get(
+                'http://localhost:8080/anything'
+            );
+        """)).fetchone()
+    assert "permission denied for table http_request_queue" in str(execinfo)
```
steve-chavez, line 92: Here's a repro of https://github.com/orgs/supabase/discussions/21450

soedirgo:
Is this ready for review? There's no code/SQL changes

steve-chavez:
Follow up on https://github.com/supabase/pg_net/pull/158

PR #133, opened 2024-07-03
fix: handle CURL and datum failures

### LINKED ISSUE #132: UNLOGGED table segfault causing whole cluster to restart
## Problem

In some rare cases, pg_net reports:

```
"background worker ""pg_net 0.8.0 worker"" (PID 132605) was terminated by signal 11: Segmentation fault",,,,,,,,,"","postmaster",,0
"terminating any other active server processes",,,,,,,,,"","postmaster",,0
"all server processes terminated; reinitializing",,,,,,,,,"","postmaster",,0
"database system was interrupted; last known up at 2024-06-03 02:15:10 UTC",,,,,,,,,"","startup",,0
"database system was not properly shut down; automatic recovery in progress",,,,,,,,,"","startup",,0
"redo starts at 0/6154CA0",,,,,,,,,"","startup",,0
```

The thing is that a segmentation fault in pg_net should only kill its worker but somehow this restarts the whole cluster.

This thread https://www.postgresql.org/message-id/flat/871rucdwkt.fsf%40alfa.kjonca indicates that this could be an UNLOGGED table bug.

So far I haven't been able to reproduce this, some Supabase tickets indicate that this can happen sometimes when invoking edge functions.

## Solution

https://github.com/supabase/pg_net/issues/62 is a possible solution. Since we would not need the UNLOGGED table for the queue.

steve-chavez:
Failures for curl_easy_setopt and curl_easy_getinfo were not handled.

Also handle NULL Datums.

Fixes #132.

### DIFF
--- a/src/worker.c
+++ b/src/worker.c
@@ -32,6 +32,25 @@
 
 #include "util.h"
 
+#define CURL_EZ_SETOPT(hdl, opt, prm) \
+  do { \
+			if (curl_easy_setopt(hdl, opt, prm) != CURLE_OK) \
+				ereport(ERROR, errmsg("Could not curl_easy_setopt(%s)", #opt)); \
+  } while (0)
+
+#define CURL_EZ_GETINFO(hdl, opt, prm) \
+  do { \
+			if (curl_easy_getinfo(hdl, opt, prm) != CURLE_OK) \
+				ereport(ERROR, errmsg("Could not curl_easy_getinfo(%s)", #opt)); \
+  } while (0)
+
+#define EREPORT_NULL_ATTR(tupIsNull, attr) \
+  do { \
+		if (tupIsNull) \
+			ereport(ERROR, errmsg("%s cannot be null", #attr)); \
+  } while (0)
+
+
 PG_MODULE_MAGIC;
 
 static char *ttl = "6 hours";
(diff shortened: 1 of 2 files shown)


src/worker.c
```diff
@@ -113,47 +132,45 @@ static CURLMcode init(CURLM *cm, char *method, char *url, int timeout_millisecon
 	(void)pushJsonbValue(&response_headers, WJB_BEGIN_OBJECT, NULL);
 	cdata->response_headers = response_headers;
 	cdata->id = id;
-	cdata->request_headers = request_headers;
-
-	request_headers = curl_slist_append(request_headers, "User-Agent: pg_net/" EXTVERSION);
```
pashkinelfe, line 118: request_headers passed to init() now not updated. Previously both cdata->request_headers and request_headers were updated.

src/worker.c
```diff
@@ -113,47 +132,45 @@ static CURLMcode init(CURLM *cm, char *method, char *url, int timeout_millisecon
...
-	curl_easy_setopt(eh, CURLOPT_HTTPHEADER, cdata->request_headers);
-	curl_easy_setopt(eh, CURLOPT_TIMEOUT_MS, timeout_milliseconds);
-	curl_easy_setopt(eh, CURLOPT_PRIVATE, cdata);
-	curl_easy_setopt(eh, CURLOPT_FOLLOWLOCATION, true);
+	CURL_EZ_SETOPT(eh, CURLOPT_WRITEFUNCTION, body_cb);
+	CURL_EZ_SETOPT(eh, CURLOPT_WRITEDATA, cdata->body);
+	CURL_EZ_SETOPT(eh, CURLOPT_HEADERFUNCTION, header_cb);
+	CURL_EZ_SETOPT(eh, CURLOPT_HEADERDATA, cdata->response_headers);
+	CURL_EZ_SETOPT(eh, CURLOPT_HEADER, 0L);
+	CURL_EZ_SETOPT(eh, CURLOPT_URL, url);
```
pashkinelfe, line 163: url should not be null (add check?)

src/worker.c
```diff
@@ -113,47 +132,45 @@ static CURLMcode init(CURLM *cm, char *method, char *url, int timeout_millisecon
 	(void)pushJsonbValue(&response_headers, WJB_BEGIN_OBJECT, NULL);
 	cdata->response_headers = response_headers;
 	cdata->id = id;
-	cdata->request_headers = request_headers;
-
-	request_headers = curl_slist_append(request_headers, "User-Agent: pg_net/" EXTVERSION);
+	cdata->request_headers = curl_slist_append(request_headers, "User-Agent: pg_net/" EXTVERSION);
 
 	if (strcasecmp(method, "GET") == 0) {
```
pashkinelfe, line 137: method should not be null (add check?)

src/worker.c
```diff
@@ -113,47 +132,45 @@ static CURLMcode init(CURLM *cm, char *method, char *url, int timeout_millisecon
 	(void)pushJsonbValue(&response_headers, WJB_BEGIN_OBJECT, NULL);
 	cdata->response_headers = response_headers;
 	cdata->id = id;
-	cdata->request_headers = request_headers;
-
-	request_headers = curl_slist_append(request_headers, "User-Agent: pg_net/" EXTVERSION);
+	cdata->request_headers = curl_slist_append(request_headers, "User-Agent: pg_net/" EXTVERSION);
```
pashkinelfe, line 135: To avoid overwriting an existing non-empty list on failure, the new list should be returned to a temporary variable which can be tested for NULL before updating the original list pointer

src/worker.c
```diff
@@ -119,43 +138,46 @@ static CURLMcode init(CURLM *cm, char *method, char *url, int timeout_millisecon
...
-	curl_easy_setopt(eh, CURLOPT_TIMEOUT_MS, timeout_milliseconds);
-	curl_easy_setopt(eh, CURLOPT_PRIVATE, cdata);
-	curl_easy_setopt(eh, CURLOPT_FOLLOWLOCATION, true);
+	CURL_EZ_SETOPT(eh, CURLOPT_WRITEFUNCTION, body_cb);
+	CURL_EZ_SETOPT(eh, CURLOPT_WRITEDATA, cdata->body);
+	CURL_EZ_SETOPT(eh, CURLOPT_HEADERFUNCTION, header_cb);
+	CURL_EZ_SETOPT(eh, CURLOPT_HEADERDATA, cdata->response_headers);
+	CURL_EZ_SETOPT(eh, CURLOPT_HEADER, 0L);
+	CURL_EZ_SETOPT(eh, CURLOPT_URL, url);
+	CURL_EZ_SETOPT(eh, CURLOPT_HTTPHEADER, cdata->request_headers);
```
pashkinelfe, line 166: Assignment on line 137 doesn't affect this code because cdata->request_headers is set at line 135. Not sure how should this be but looks as 137 and 135 should be in reverse order. Also curl_slist_append at line 137 could output null at error, thus deleting the whole list instead of adding an element.

steve-chavez: Addressed both concerns. `curl_slist_append` is now tested for NULL.

pashkinelfe (approved):

Issue #106, opened 2023-09-18
Multiple versions of nixpkgs


steve-chavez:
Currently we use [multiple versions of nixpkgs](https://github.com/supabase/pg_net/blob/9d3f5be9b90663fb1a6ba92df61698236e399363/shell.nix#L4-L18). There are some issues that prevents us from using only the newest nixpkgs.

https://github.com/supabase/pg_net/blob/9d3f5be9b90663fb1a6ba92df61698236e399363/shell.nix#L4-L18

## Tasks

- [x] newest nixpkgs is only python3
  - many tests give the `sqlalchemy.exc.ArgumentError: Textual SQL expression` error
  - cleared on https://github.com/supabase/pg_net/pull/107
- [ ] pathod command no longer exists https://github.com/mitmproxy/mitmproxy/issues/4273#issuecomment-1724490887. Ideally the maintainers reply with a way forward using `mitmproxy`.
- [x] newest openresty doesn't have a prefix option: https://github.com/supabase/pg_net/blob/9d3f5be9b90663fb1a6ba92df61698236e399363/shell.nix#L2-L3

This is not a big problem since we can keep using multiple versions. It only adds some extra time for building/testing - not too much time (only download time, not compile time) because https://cache.nixos.org/  is already populated.

steve-chavez:
> pathod command no longer exists https://github.com/mitmproxy/mitmproxy/issues/4273#issuecomment-1724490887. Ideally the maintainers reply with a way forward using mitmproxy.

For the above the way forward is to have the same functionality on https://github.com/steve-chavez/ngx_pathological, then we can just use Nginx for the tests.

PR #136, opened 2024-08-07
Fix malformed headers

steve-chavez:
When having a malformed header such as:

```
$ curl localhost:8080/pathological?malformed-header=header-injection -i
HTTP/1.1 200 OK
Server: nginx/1.26.0

HeaderInjection
Injected-Header: This header contains an injection
```

The background worker crashes.

To prove the above, we use https://github.com/steve-chavez/ngx_pathological.

### DIFF
--- a/src/worker.c
+++ b/src/worker.c
@@ -37,7 +37,6 @@ PGDLLEXPORT void pg_net_worker(Datum main_arg) pg_attribute_noreturn();
 typedef struct {
   int64 id;
   StringInfo body;
-  JsonbParseState* response_headers;
   struct curl_slist* request_headers;
 } CurlData;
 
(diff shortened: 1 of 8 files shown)


steve-chavez:
Tempted to use this newish (2022) curl feature for parsing the headers: https://curl.se/libcurl/c/curl_easy_nextheader.html ([announcement](https://daniel.haxx.se/blog/2022/03/22/a-headers-api-for-libcurl/)).  It's pretty convenient.

It's available from curl 7.83, I believe now that we have Nix on prod it should be doable to get this version.

Note: We need to document this requirement on the README.

steve-chavez:
Some failure about GLIBC on CI https://github.com/supabase/pg_net/actions/runs/10296126418/job/28496960251?pr=136

```
waiting for server to start....2024-08-08 04:53:06.180 UTC [13164] FATAL:  could not load library "/nix/store/hcpahbjn22r4ddbqgsh0pfmgy5vy7lls-postgresql-and-plugins-12.5/lib/pg_net.so": 
/nix/store/1yvpgm763b3hvg8q4fzpzmflr5674x4j-glibc-2.32-10/lib/libc.so.6: version `GLIBC_2.33' not found (required by /nix/store/36kr91f8s509z2r8hqvpayw5bqyqgyzr-curl-8.7.1/lib/libcurl.so.4)
```

Easiest way to solve it is to upgrade to a new nixpkgs version. A bit related to https://github.com/supabase/pg_net/issues/106

soedirgo (approved):

src/worker.c
```diff
@@ -430,16 +423,14 @@ _PG_init(void)
...
-  worker.bgw_flags = BGWORKER_SHMEM_ACCESS | BGWORKER_BACKEND_DATABASE_CONNECTION;
-  worker.bgw_start_time = BgWorkerStart_RecoveryFinished;
-  snprintf(worker.bgw_library_name, BGW_MAXLEN, "pg_net");
-  snprintf(worker.bgw_function_name, BGW_MAXLEN, "pg_net_worker");
-  snprintf(worker.bgw_name, BGW_MAXLEN, "pg_net " EXTVERSION " worker");
-  worker.bgw_restart_time = 1;
-  worker.bgw_main_arg = (Datum) 0;
-  worker.bgw_notify_pid = 0;
-  RegisterBackgroundWorker(&worker);
+  RegisterBackgroundWorker(&(BackgroundWorker){
```
pashkinelfe, line 426: IMO previous state of pg_init() was completely ok and also was easier to read.

steve-chavez: The new way using C99 designated initializer requires fewer LOC and avoids the needed to `memset` to 0. Readability is highly subjective but fewer LOCs in this case is a win IMO.

PR #138, opened 2024-08-09
bump to 0.9.3

steve-chavez:


samrose (approved):

steve-chavez:
All release artifacts failed after this https://github.com/supabase/pg_net/actions/runs/10322326776/job/28577307445.

That makes sense because they use an older libcurl library.

Issue #86, opened 2023-01-10
Webhooks fail when multiple inserts are done


devdras:
# Bug report

## Describe the bug

I have a webhook configured to watch a table for inserts. On an insert a POST request should be fired to an API hosted on Vercel. 
This works when one insert operation is carried out. If multiple inserts are carried out: .insert([{message: 1}, {message: 2}, {message: 3}]) then the webhooks aren't fired at all. If instead we carry out multiple inserts by calling the Supabase API multiple times then the webhook is fired for only some of the inserts. 

## To Reproduce

Steps to reproduce the behavior, please provide code snippets or a repository:

1. Configure a webhook on a table to watch for insert operations and call an external API
2. Use the Supabase JS library to perform multiple inserts like: `let { data, error } = await supabase
              .from("example_table")
              .insert([{message: 1}, {message: 2}, {message: 3}, {message: 4}, {message: 5}, {message: 6}, etc ])
              .select("message");`

4. Check logs for external API to confirm it didn't receive any requests
5. See error

Something to note is that I was inserting 32 things at once.

## Expected behavior

For each insert, I would expect the external API to be called. It would also help if there was a log in Supabase that showed what happens when the webhook is run.

## System information

- OS: N/A
- Browser (if applies): N/A
- Version of supabase-js: 2.0.0-rc.8
- Version of Node.js: v16.17.0 on desktop, 18.x on Vercel (per their docs)

steve-chavez:
We've tested pg_net sending 1K inserts per second, this to a local nginx server though.

@devdras @danrasmuson Which webhook service are you using? Likely it's rate limiting the requests.

danrasmuson:
> We've tested pg_net sending 1K inserts per second, this to a local nginx server though.
> 
> 
> 
> @devdras @danrasmuson Which webhook service are you using? Likely it's rate limiting the requests.

Im using vercel with serverless functions

matt-aitken:
We're experiencing the same problem. When inserting records in sequence all the webhooks are received. But if we insert 30 rows in parallel none of the webhooks send and the `net/_http_response` table is full of `Timeout was reached` errors.

steve-chavez:
@matt-aitken Are you using vercel serverless functions too?

lohrmann:
@steve-chavez We're able to reliably reproduce it based on the number of new entries that get bulk added to the table—have been in touch with your support team about this as well, but unable to resolve it. 
- The webhook trigger is calling an AWS Lambda function
[...]

PR #139, opened 2024-08-15
feat: epoll based worker

steve-chavez:
This replaces the usage of [curl_multi_perform](https://curl.se/libcurl/c/curl_multi_perform.html) (which uses `select()`) for [curl_multi_socket_action](https://curl.se/libcurl/c/curl_multi_socket_action.html) + `epoll()`, which is more performant (requests finish in less time) and more reliable (with 0 timeouts). 

It solves #86. Using the cloud test added on https://github.com/supabase/pg_net/pull/150, on this branch we get the following results:

```
postgres=# call repro_timeouts(10000);
NOTICE:  Waiting until 10000 requests complete
NOTICE:  Stats: {"request_successes":10000,"request_failures":0,"last_failure_error":null}
NOTICE:  Time taken: 00:01:01.360567
CALL

postgres=# call repro_timeouts(20000);
NOTICE:  Waiting until 20000 requests complete
NOTICE:  Stats: {"request_successes":20000,"request_failures":0,"last_failure_error":null}
NOTICE:  Time taken: 00:01:44.942479

postgres=# call repro_timeouts(100000);
NOTICE:  Waiting until 100000 requests complete
NOTICE:  Stats: {"request_successes":100000,"request_failures":0,"last_failure_error":null}
NOTICE:  Time taken: 00:08:58.325675

postgres=# call repro_timeouts (1000000);
NOTICE:  Waiting until 1000000 requests complete
NOTICE:  Stats: {"request_successes":1000000,"request_failures":0,"last_failure_error":null}
NOTICE:  Time taken: 01:45:33.085331
CALL
```

On master:

```
postgres=# call repro_timeouts(10000);
NOTICE:  Waiting until 10000 requests complete
NOTICE:  Stats: {"request_successes":8886,"request_failures":1114,"last_failure_error":"Ti
meout was reached"}
NOTICE:  Time taken: 00:01:44.061517
CALL
postgres=# call repro_timeouts(20000);
NOTICE:  Waiting until 20000 requests complete
NOTICE:  Stats: {"request_successes":18071,"request_failures":1929,"last_failure_error":"T
imeout was reached"}
NOTICE:  Time taken: 00:04:05.716657
CALL

postgres=# call repro_timeouts(30000);
NOTICE:  Waiting until 30000 requests complete
NOTICE:  Stats: {"request_successes":26939,"request_failures":3061,"last_failure_error":"T
imeout was reached"}
NOTICE:  Time taken: 00:06:31.642828
CALL
```

[...]

PR #150, opened 2024-09-17
test: add cloud test for reproducing timeouts

steve-chavez:
This is a test that can only be ran manually since it requires AWS. It adds NixOps wrapper scripts to reproduce https://github.com/supabase/pg_net/issues/86.

The steps are:

```
$ net-cloud-deploy
$ net-cloud-reproduce-timeouts
NOTICE:  Waiting until 10000 requests complete
NOTICE:  Stats: {"request_successes":8882,"request_failures":1118,"last_failure_error":"Timeout was reached"}
NOTICE:  Time taken: 00:01:44.341157
```

Then destroy the cloud setup with `net-cloud-destroy`

Note that it takes 01:44, it should be ~50 seconds to finish processing all the 10K requests, considering batch_size=200 (the default).

### DIFF
--- a/nix/nixops.nix
+++ b/nix/nixops.nix
@@ -115,16 +115,44 @@ in {
       initialScript = pkgs.writeText "init-sql-script" ''
         create extension pg_net;
 
-        alter system set pg_net.batch_size to 32000;
-
-        select net.worker_restart();
-
         create view pg_net_stats as
         select
           count(*) filter (where error_msg is null) as request_successes,
           count(*) filter (where error_msg is not null) as request_failures,
           (select error_msg from net._http_response where error_msg is not null order by id desc limit 1) as last_failure_error
         from net._http_response;
+
+        create or replace procedure repro_timeouts(number_of_requests int default 10000, url text default 'http://server/post') as ''$''$
+        declare
+          last_id bigint;
+          first_time timestamptz;
+          second_time timestamptz;
+          time_taken interval;
+        begin
+          delete from net._http_response;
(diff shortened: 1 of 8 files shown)


steve-chavez:
As a follow up note, I can't reproduce this on the https://github.com/supabase/pg_net/pull/139 branch.

I believe this is due to `select` (which `curl_multi_perform` uses) vs `epoll` performance. There's some internal extra time required for getting all responses.

nix/nixops.nix
```diff
@@ -136,22 +132,48 @@ in {
...
+          psql -U postgres -c "alter system reset pg_net.batch_size;"
+          psql -U postgres -c "select net.worker_restart();"
+        ''
+      )
+      (
+        pkgs.writeShellScriptBin "psql-reproduce-timeouts" ''
+          set -euo pipefail
+
+          number_of_requests=10000
+          sleep_time=60 # Using the default batch_size (200), this is enough time to finish the whole $number_of_requests
```
steve-chavez, line 169: Considering the default batch_size (200), we should be able to wait ~50 seconds to get all the 10K responses. However this is not the case, most likely due to the `select()` overhead as mentioned on https://github.com/supabase/pg_net/pull/150#issuecomment-2354262406.

samrose (approved):

steve-chavez:
Modified the script to now include the actual time taken to process all requests. See [above](https://github.com/supabase/pg_net/pull/150#issue-2529715130)

PR #156, opened 2024-10-10
refactor: use pg WaitEvent API instead of epoll

steve-chavez:
This allows compiling on other platforms such as MacOS. (problem previously mentioned on https://github.com/supabase/pg_net/pull/139#issuecomment-2387312836)

It mostly works at this stage but with some limitations:

- [ ] Doing 10K requests fail with "Bad file descriptor"
  + I believe this is because we're not deleting file descriptors since the WaitEvent API doesn't provide EPOLL_CTL_DEL. It only provides EPOLL_CTL_ADD/MOD
- [ ] timerfd is still Linux specific and pg provides no abstraction over it.

### DIFF
--- a/src/core.c
+++ b/src/core.c
@@ -24,7 +24,6 @@
 #include <curl/multi.h>
 
 #include <sys/epoll.h>
-#include <sys/timerfd.h>
 #include <unistd.h>
 #include <errno.h>
 #include <string.h>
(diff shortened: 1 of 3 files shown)


steve-chavez:
Using a simpler approach for now on https://github.com/supabase/pg_net/pull/165.

PR #158, opened 2024-10-17
fix: permission denied for new roles and remove security definer functions

steve-chavez:
Closes https://github.com/supabase/pg_net/issues/72. Also addresses https://github.com/orgs/supabase/discussions/21450.

### DIFF
--- a/test/test_privileges.py
+++ b/test/test_privileges.py
@@ -63,6 +63,17 @@ def test_net_on_another_role(sess):
     ).fetchone()
     assert response[0] == "SUCCESS"
 
+    ## can use the net.worker_restart function
+    response = sess.execute(
+        text(
+            """
+        set local role to another;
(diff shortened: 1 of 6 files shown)


test/test_privileges.py
```diff
@@ -0,0 +1,80 @@
...
+        select net.worker_restart();
+    """
+        )
+    ).fetchone()
+    assert response[0] == True
+
+    sess.execute(text("""
+        set local role postgres;
+        drop role another;
+    """))
```
steve-chavez, line 80: Note that this changes are tested

steve-chavez:
After this is merged, we can integrate the `supabase_functions` as a dependent extension here (using https://github.com/supabase/postgres/pull/1152) and solve all the permission issues.

soedirgo (approved):

PR #168, opened 2024-11-08
feat: set appname for pg_stat_activity entry

steve-chavez:
This is so `select * from pg_stat_activity where application_name like '%pg_net%';` can work as usual.

### DIFF
--- a/src/event.c
+++ b/src/event.c
@@ -14,7 +14,7 @@ typedef struct epoll_event epoll_event;
 typedef struct itimerspec itimerspec;
 
 int inline wait_event(int fd, event *events, size_t maxevents, int wait_milliseconds){
-  return epoll_wait(fd, events, maxevents, /*timeout=*/1000);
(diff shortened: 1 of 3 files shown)


soedirgo (approved):

za-arthur (approved):
Thank you! Looks good to me.
It is possible to get status of the worker by checking `backend_type`:

```
=# select datname, application_name, state, backend_type from pg_stat_activity where backend_type like 'pg_net%';
   datname    │ application_name │ state  │     backend_type
──────────────┼──────────────────┼────────┼──────────────────────
 pre_existing │                  │ (null) │ pg_net 0.12.0 worker
```

But I suppose altering `application_name` could make it more friendlier to users.

steve-chavez:
> But I suppose altering application_name could make it more friendlier to users.

Yes, this was just a nicety. I noticed pg_cron does the same.

Issue #169, opened 2024-11-26
No way to know the underlying cause of timeouts


steve-chavez:
## Problem

If a timeout occurs, we only get a `Timeout was reached` error:

```
 id | status_code | content_type | headers | content | timed_out |      error_msg      |            created            
----+-------------+--------------+---------+---------+-----------+---------------------+-------------------------------
  1 |             |              |         |         |           | Timeout was reached | 2024-11-26 13:17:11.479181-05
(1 row)
```

But a timeout might occur because the connection takes too much time (network latency) or because the server response is slow. 

## Solution

Curl provides different timings for the request/response duration:

```
time_appconnect
             The time, in seconds, it took from the start until the SSL/SSH/etc connect/handshake to the remote host was completed.

time_connect   The time, in seconds, it took from the start until the TCP connect to the remote host (or proxy) was completed.

time_namelookup
             The time, in seconds, it took from the start until the name resolving was completed.

time_pretransfer
             The time, in seconds, it took from the start until the file transfer was just about to begin. This includes all  pre-transfer  commands  and  negotiations
             that are specific to the particular protocol(s) involved.

time_redirect  The time, in seconds, it took for all redirection steps including name lookup, connect, pretransfer and transfer before the final transaction was started.
             time_redirect shows the complete execution time for multiple redirections.

time_starttransfer
             The time, in seconds, it took from the start until the first byte was just about to be transferred. This includes time_pretransfer and also the  time  the
             server needed to calculate the result.

time_total     The total time, in seconds, that the full operation lasted.
```

We can include these details in the error message.

steve-chavez:
Adding a TCP handshake delay (with [tc](https://www.google.com/search?q=tc+linux&oq=tc+linux&gs_lcrp=EgZjaHJvbWUqBggAEEUYOzIGCAAQRRg7MgYIARBFGDwyBggCEEUYPDIGCAMQRRg8MgYIBBAuGEDSAQc4ODBqMGoxqAIAsAIA&sourceid=chrome&ie=UTF-8)), plus an http delay ([ref](https://github.com/steve-chavez/ngx_pathological?tab=readme-ov-file#response-delay)), prints the following stats:

```
$ sudo tc qdisc add dev lo root netem delay 500ms

[...]

PR #170, opened 2024-12-04
feat: add detailed timeout error

steve-chavez:
Closes https://github.com/supabase/pg_net/issues/169.

When a timeout occurs, now the error message is like:

```
Timeout of 1000 ms reached. Total time: 1000.527000 ms (DNS time: 0.084000 ms, TCP/SSL handshake time: 0.314000 ms, HTTP Request/Response time: 1000.014000 ms)
```

Before it used to be:

```
Timeout was reached
```

### DIFF
--- a/src/core.c
+++ b/src/core.c
@@ -36,7 +36,12 @@ typedef struct {
   int64 id;
   StringInfo body;
   struct curl_slist* request_headers;
+  int32 timeout_milliseconds;
 } CurlData;
+
+typedef struct {
+  char msg[256];
+} curl_timeout_msg;
 
 static size_t
 body_cb(void *contents, size_t size, size_t nmemb, void *userp)
@@ -54,6 +59,8 @@ static void init_curl_handle(CURLM *curl_mhandle, MemoryContext curl_memctx, int
   CurlData *cdata = palloc(sizeof(CurlData));
   cdata->id   = id;
   cdata->body = makeStringInfo();
+
+  cdata->timeout_milliseconds = timeout_milliseconds;
 
   if (!headersBin.isnull) {
     ArrayType *pgHeaders = DatumGetArrayTypeP(headersBin.value);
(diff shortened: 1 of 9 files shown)


soedirgo (approved):

src/core.c
```diff
@@ -164,16 +171,101 @@ void delete_expired_responses(char *ttl, int batch_size){
...
+
+  // convert seconds to milliseconds
+  double dns_time_ms       = _dns_time       * 1000;
+  double handshake_time_ms = _handshake_time * 1000;
+  double http_time_ms      = _http_time      * 1000;
+  double total_time_ms     = total           * 1000;
+
+  // build the error message
+  curl_timeout_msg result = {.msg = {}};
+  sprintf(result.msg,
```
za-arthur, line 245: It is safer to call `snprintf` to make sure that the result string won't be larger than `result.msg`.

za-arthur (commented):
Thank you for the PR. All tests pass on my laptop, although I had to increase timeout in `time.sleep` call within `test_http_detailed_timeout()` and `test_http_get_succeed_with_gt_timeout()`.

I have couple of suggestions. One is mentioned below. And the second: would it make more sense to add new column `error_detail` to `net._http_response`? That way it would be easier to make queries to the table without using pattern matching.

src/core.c
steve-chavez: Good call, done. Also went ahead and moved all this logic to a more proper `errors.hs` module.

steve-chavez:
> would it make more sense to add new column error_detail to net._http_response? That way it would be easier to make queries to the table without using pattern matching.

@za-arthur I thought about that too, but the problem is that users will likely ignore the new detail column and the goal is for them to pay attention to these timings.

Additionally, we don't add more details to other error messages, just this one. All the other errors come raw from libcurl. So not much use for the new column.

za-arthur (approved):
Thank you for the fix. LGTM.

Issue #77, opened 2023-01-06
expand pg_net functionality with more operations and other data types


davidaventimiglia:
# Feature request

expand pg_net functionality with more operations and other data types

## Is your feature request related to a problem? Please describe.

Yes.  I would like to use `pg_net` to access REST micro-services in an asynchronous manner, when those services rely on other HTTP methods besides just `GET`, such as `PUT`, `PATCH`, and `DELETE`.  Moreover, sometimes those services work with payloads that are not JSON and therefore cannot be passed to a PostgreSQL function as a `json` or `jsonb` data type.

## Describe the solution you'd like

In addition to the existing `net.http_get(url text, params jsonb, headers jsonb, timeout_milliseconds int)` and `net.http_post(url text, params jsonb, headers jsonb, timeout_milliseconds int)` functions, I would like for there to be a master function `net.http(request http_request, timeout_milliseconds int)` function similar to the `http.http(request http_request)` function in the [psql-http](https://github.com/pramsey/pgsql-http) extension.  Like in that extension, `http_request` would be a data type that has both a `method` and a `content` attribute, the latter being `varchar`.  This would be enough to support other HTTP methods and other payloads.

## Describe alternatives you've considered

I have considered and even used the synchronous `http` extension in conjunction with custom tables and the `pg_cron` extension to (re)implement a pseudo-async processor, but it's cumbersome and duplicative of the work that's in the `pg_net` extension.

## Additional context

No other context is relevant.

tvogel:
Lack of `DELETE` makes it impossible to use `pg_net` to clean up supabase storage objects in `UPDATE`/`DELETE` Postgres triggers. Or am I missing something?

JulienLecoq:
> Lack of `DELETE` makes it impossible to use `pg_net` to clean up supabase storage objects in `UPDATE`/`DELETE` Postgres triggers. Or am I missing something?

Was thinking the same

steve-chavez:
DELETE is now supported, see https://github.com/supabase/pg_net/issues/63#issuecomment-1402780864.

> Can be used on v0.7 as:

        select net.http_delete(
            url:='https://httpbin.org/delete'
        ,   params:= '{"param-foo": "bar"}'
        ,   headers:= '{"X-Baz": "foo"}'
        );

erik-beus:
> DELETE is now supported, see https://github.com/supabase/pg_net/issues/63#issuecomment-1402780864.
> 
> 
> 
[...]

PR #173, opened 2024-12-12
feat: support net.http_delete with request body

steve-chavez:
Addresses https://github.com/supabase/pg_net/issues/77#issuecomment-1465029556. Now it's possible to do:

```sql
select net.http_delete(
    url  :='http://localhost:8080/delete_w_body'
,   body := '{"key": "val"}'
);
```

Backwards compatibility is tested. No body will be sent if the body parameter is NULL (the default).

## Notes

- I tried adding an overloaded function for `net.http_delete` instead of modifying the existing one, but that caused breakage since every call like:

```sql
        select net.http_delete(
            url:='http://localhost:8080/delete'
        ,   params:= '{"param-foo": "bar"}'
        ,   headers:= '{"X-Baz": "foo"}'
        );
```

Failed with an error related to ambiguity (two functions with the same named parameters).

### DIFF
--- a/nix/nginx/conf/custom.conf
+++ b/nix/nginx/conf/custom.conf
@@ -49,6 +49,15 @@ location /delete {
   echo_duplicate 1 $echo_client_request_headers$is_args$query_string;
 }
 
+location /delete_w_body {
+  if ($request_method != 'DELETE'){
+      return 405;
+  }
+
+  echo_read_request_body;
+  echo $request_body;
+}
+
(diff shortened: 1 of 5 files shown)


soedirgo:
Let's only merge this after a migration file is added, otherwise pg_net.sql and the upgrade scripts go out of sync

steve-chavez:
Ok then. I've added the migration file.

soedirgo (approved):

Issue #164, opened 2024-10-26
Frequent queries pollute `pg_stat_statements` 


steve-chavez:
## Problem

Currently we run two queries every second ([ref1](https://github.com/supabase/pg_net/blob/1eb33d508bfc67bc7a64138142042545fc4e4650/src/worker.c#L161-L163), [ref2](https://github.com/supabase/pg_net/blob/1eb33d508bfc67bc7a64138142042545fc4e4650/src/core.c#L201-L223), [ref3](https://github.com/supabase/pg_net/blob/1eb33d508bfc67bc7a64138142042545fc4e4650/src/core.c#L291-L310)), note that one does a DELETE to avoid having the `http_request_queue` from growing too much. This pollutes `pg_stat_statements` and causes some users to complain and wrongly think that pg_net is causing a lot of iops.

## Solution 1

We can instead add the requests done with `net.http_get` and others to an in-memory queue, and then read from this queue whenever there are entries available.

This can be done without a breaking change by adding a trigger to the `net.http_request_queue` table.

## Solution 2 (preferred)

A simpler change would be just waking up the worker whenever there's an `http_get` or other function call and put it to sleep whenever there are 0 rows processed.

This would mean migrating the `net.http_<method>` functions to use C instead of SQL. 

## Related

Solving this issue can be a stepping stone towards https://github.com/supabase/pg_net/issues/62.

PR #182, opened 2025-04-16
repro memory leak under pg_stat_monitor

steve-chavez:
## Problem

When adding `pg_stat_monitor` like:

```
shared_preload_libraries='pg_net, pg_stat_monitor'
```

Memory usage keeps increasing for the pg_net background worker.

To prove this execute in one terminal:

```bash
xpg -v 15 psql
```

And this in another:

```bash
watch -n 1 'ps -p $(cat build-15/bgworker.pid) -o %mem,rss,vsz,comm'

%MEM   RSS    VSZ COMMAND
 0.1 19164 722508 postgres
```

The RSS should be seen as always growing, which proves the memory leak.

## Root cause

pg_stat_monitor statistics are split into configurable time intervals ([buckets](https://docs.percona.com/pg-stat-monitor/user_guide.html#time-buckets)). As one bucket fills or a time interval passes, pg_stat_monitor creates a new one, keeping historical data in distinct buckets.

pg_net queries its internal tables each second.

This means that the "memory leak" is actually expected behavior by pg_stat_monitor.

We need to fix this in pg_net itself, by doing https://github.com/supabase/pg_net/issues/164.

### DIFF
--- a/nix/pg_stat_monitor.nix
+++ b/nix/pg_stat_monitor.nix
@@ -1,0 +1,32 @@
+{ lib, stdenv, fetchFromGitHub, postgresql }:
+
+stdenv.mkDerivation rec {
+  pname = "pg_stat_monitor";
+  version = "2.1.1";
+
+  buildInputs = [ postgresql ];
+
+  src = fetchFromGitHub {
+    owner = "percona";
+    repo = pname;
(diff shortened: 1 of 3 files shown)


steve-chavez:
Can confirm this no longer happens. Closing.

Issue #200, opened 2025-06-27
response insertion can fail when dropping extension


steve-chavez:
## Problem

This can happen when having a slow request, and then dropping extension:

```psql
postgres=# select net.http_get(url := 'http://localhost:8080/pathological?status=200&delay=10');
 http_get 
----------
        1
(1 row)

2025-06-26 20:23:05.978 -05 [251232] DEBUG:  Deleted 0 expired rows
2025-06-26 20:23:05.978 -05 [251232] DEBUG:  Consumed 1 request rows
2025-06-26 20:23:05.979 -05 [251232] DEBUG:  Pending curl running_handles: 1
2025-06-26 20:23:05.979 -05 [251232] DEBUG:  Pending curl running_handles: 1
2025-06-26 20:23:05.979 -05 [251232] DEBUG:  Pending curl running_handles: 1
2025-06-26 20:23:06.179 -05 [251232] DEBUG:  Pending curl running_handles: 1
2025-06-26 20:23:07.182 -05 [251232] DEBUG:  Pending curl running_handles: 1

postgres=#   drop extension pg_net cascade;
DROP EXTENSION

2025-06-26 20:23:09.191 -05 [251232] DEBUG:  Pending curl running_handles: 1
2025-06-26 20:23:10.196 -05 [251232] DEBUG:  Pending curl running_handles: 1
2025-06-26 20:23:10.980 -05 [251232] DEBUG:  relation "net._http_response" does not exist at character 19
2025-06-26 20:23:10.980 -05 [251232] QUERY:        insert into net._http_response(id, error_msg) values ($1, $2)
2025-06-26 20:23:10.980 -05 [251232] ERROR:  relation "net._http_response" does not exist at character 19
2025-06-26 20:23:10.980 -05 [251232] QUERY:        insert into net._http_response(id, error_msg) values ($1, $2)
2025-06-26 20:23:10.984 -05 [251226] LOG:  background worker "pg_net 0.17.0 worker" (PID 251232) exited with exit code 1
2025-06-26 20:23:11.990 -05 [251226] DEBUG:  starting background worker process "pg_net 0.17.0 worker"
2025-06-26 20:23:11.992 -05 [251245] PANIC:  latch already owned by PID 251232
2025-06-26 20:23:12.091 -05 [251226] LOG:  background worker "pg_net 0.17.0 worker" (PID 251245) was terminated by signal 6: Aborted
[...]

PR #201, opened 2025-06-27
fix: don't crash when dropping extension

steve-chavez:
Closes https://github.com/supabase/pg_net/issues/200

### DIFF
--- a/src/worker.c
+++ b/src/worker.c
@@ -173,55 +173,59 @@ void pg_net_worker(__attribute__ ((unused)) Datum main_arg) {
 
     elog(DEBUG1, "Deleted %zu expired rows", expired_responses);
 
+    StartTransactionCommand();
+    PushActiveSnapshot(GetTransactionSnapshot());
+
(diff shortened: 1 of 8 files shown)


src/worker.c
```diff
@@ -173,55 +173,59 @@ void pg_net_worker(__attribute__ ((unused)) Datum main_arg) {
...
-        insert_curl_responses(&lstate, CurlMemContext);
-      }
 
-      elog(DEBUG1, "Pending curl running_handles: %d", running_handles);
-    } while (running_handles > 0); // run while there are curl handles, some won't finish in a single iteration since they could be slow and waiting for a timeout
+        elog(DEBUG1, "Pending curl running_handles: %d", running_handles);
+      } while (running_handles > 0); // run while there are curl handles, some won't finish in a single iteration since they could be slow and waiting for a timeout
+    }
 
-    pfree(events);
```
steve-chavez, line 224: This is unnecessary now as we only use stack allocation for the events array.

laurenceisla (approved):

PR #221, opened 2025-08-01
refactor: use MyProc->procLatch instead of new one

steve-chavez:
`&MyProc->procLatch` is already a shared latch, so reuse that one instead of creating a new shared latch.


za-arthur (approved):
Thank you for the PR!

I tested and re-checked that `DROP DATABASE ...` and `SELECT net.wake()` commands still work.

Issue #220, opened 2025-08-01
Correctly initialize shared memory


za-arthur:
# Bug report

- [ ] I confirm this is a bug with Supabase, not with my own application. 
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

Extensions should use `shmem_request_hook` to request extra shared memory. For example:
https://github.com/postgres/postgres/blob/REL_17_5/contrib/pg_stat_statements/pg_stat_statements.c#L488-L495
I didn't find if `pg_net` uses that hook or requests shared memory in another way.

The hook `shmem_startup_hook` should acquire `AddinShmemInitLock` when calling `ShmemInitStruct` to avoid race conditions. For example:
https://github.com/postgres/postgres/blob/REL_17_5/contrib/pg_stat_statements/pg_stat_statements.c#L527-L554
`pg_net` [doesn't do that](https://github.com/supabase/pg_net/blob/v0.19.4/src/worker.c#L361-L379).

You also can see more information in the documentation:
https://www.postgresql.org/docs/current/xfunc-c.html#XFUNC-SHARED-ADDIN

steve-chavez:
Many thanks, this is an oversight indeed. Will fix.

PR #222, opened 2025-08-04
fix: incorrect shared memory setup 

steve-chavez:
Before the shared memory wasn't guarded with lwlocks plus there was no call to RequestAddinShmemSpace to estimate used memory (the latter only available on pg >= 15).

Closes https://github.com/supabase/pg_net/issues/220

### DIFF
--- a/src/worker.c
+++ b/src/worker.c
@@ -37,6 +37,11 @@ static int                      guc_batch_size;
 static char*                    guc_database_name;
 static char*                    guc_username;
 static MemoryContext            CurlMemContext = NULL;
+
+#if PG15_GTE
+static shmem_request_hook_type  prev_shmem_request_hook = NULL;
+#endif
+
 static shmem_startup_hook_type  prev_shmem_startup_hook = NULL;
 static volatile sig_atomic_t    got_sighup = false;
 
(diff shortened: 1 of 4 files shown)


src/worker.c
```diff
@@ -357,12 +362,28 @@ void pg_net_worker(__attribute__ ((unused)) Datum main_arg) {
   proc_exit(EXIT_FAILURE);
 }
 
+#if PG15_GTE
+static Size net_memsize(void) {
+  return MAXALIGN(sizeof(WorkerState));
+}
+
+static void net_shmem_request(void) {
```
za-arthur, line 370:
```suggestion
static Size net_memsize(void) {
  return MAXALIGN(sizeof(WorkerState));
}

#if PG15_GTE
static void net_shmem_request(void) {
```

src/worker.c
```diff
@@ -357,12 +362,28 @@ void pg_net_worker(__attribute__ ((unused)) Datum main_arg) {
...
+  if (prev_shmem_request_hook)
+    prev_shmem_request_hook();
+
+  RequestAddinShmemSpace(net_memsize());
+  RequestNamedLWLockTranche("pg_net", 1);
```
za-arthur, line 375: I think `pg_net` doesn't need to call `RequestNamedLWLockTranche()` since it doesn't have its own LWLocks and for example doesn't call `GetNamedLWLockTranche()` during startup.

za-arthur:
```suggestion
```

src/worker.c
```diff
@@ -397,6 +420,11 @@ void _PG_init(void) {
     .bgw_restart_time = net_worker_restart_time_sec,
   });
 
+#if PG15_GTE
+  prev_shmem_request_hook = shmem_request_hook;
+  shmem_request_hook = net_shmem_request;
+#endif
```
za-arthur, line 426:
```suggestion
#if PG15_GTE
  prev_shmem_request_hook = shmem_request_hook;
  shmem_request_hook = net_shmem_request;
#else
  RequestAddinShmemSpace(net_memsize());
#endif
```

za-arthur (changes requested):
> Before the shared memory wasn't guarded with lwlocks plus there was no call to RequestAddinShmemSpace to estimate used memory (the latter only available on pg >= 15).

Thank you for the PR! I think `RequestAddinShmemSpace()` itself is available before Postgres 15, but indeed `shmem_request_hook` hook was added on Postgres 15 only. Before 15 extensions were requesting memory within `_PG_init()` typically. See:

https://github.com/postgres/postgres/blob/REL_14_18/contrib/pg_prewarm/autoprewarm.c#L141

And here is the commit which changed it:

https://github.com/postgres/postgres/commit/4f2400cb3f10aa79f99fba680c198237da28dd38#diff-ffc42154e133dc207f731ed03bda00d70946ec7533ed67447b8e143e32c059dfR142-R145

I added this suggestion below.

Issue #216, opened 2025-07-16
High Memory usage when the `net._http_response` table is too big


steve-chavez:
## Problem

When having:

```
postgres=# \dt+ net.*
                                             List of relations
 Schema |        Name        | Type  |     Owner      | Persistence | Access method |  Size   | Description
--------+--------------------+-------+----------------+-------------+---------------+---------+-------------
 net    | _http_response     | table | supabase_admin | unlogged    | heap          | 1104 MB |
 net    | http_request_queue | table | supabase_admin | unlogged    | heap          | 464 kB  |
(2 rows)

postgres=# select count(*) from net._http_response;
 count
--------
 442627
(1 row)
```

The `pg_net` worker consumes a lot of memory:

```
ps -e -o pid,ppid,cmd,%mem,%cpu --sort=-%mem | grep postgres
    948     787 postgres: main: pg_net 0.13 55.0  0.5
```

The `pg_net.ttl` is respected and working. This is likely because the query we execute:

```sql
    WITH\
    rows AS (\
      SELECT ctid\
      FROM net._http_response\
      WHERE created < now() - $1\
      ORDER BY created\
      LIMIT $2\
    )\
    DELETE FROM net._http_response r\
    USING rows WHERE r.ctid = rows.ctid",
```

Uses an `ORDER BY` which is not indexed.

This happens on pg_net 0.13.0.

## Solution

Reproduce the problem and check if an index on `_http_response.created` solves it.

PR #223, opened 2025-08-14
refactor: use curl_global_init_mem with pg allocators

steve-chavez:
Uses `curl_global_init_mem(CURL_GLOBAL_DEFAULT, net_malloc, net_free, net_realloc, net_strdup, net_calloc)` to keep better track of memory allocation by curl.

The aim is to fix https://github.com/supabase/pg_net/issues/216.

Unfortunately this currently segfaults when reaching 2 iterations:

```sql
-- do twice
select net.http_get('http://localhost:8080/pathological?status=200') from generate_series(1,10);
select net.http_get('http://localhost:8080/pathological?status=200') from generate_series(1,10);

-- or once (this will do two iterations since the batch size defaults to 200)
select net.http_get('http://localhost:8080/pathological?status=200') from generate_series(1,300);
```

The logs:

```
2025-08-14 17:55:08.571 -05 [643019] DEBUG:  net_realloc(palloc): a((nil)), sz(32)
2025-08-14 17:55:08.571 -05 [643019] DEBUG:  net_malloc: sz(32)
2025-08-14 17:55:08.665 -05 [643013] LOG:  background worker "pg_net 0.19.5 worker" (PID 643019) was terminated by signal 11: Segmentation fault
```

Using `gdb` (via convenience `sudo xpg gdb`) will show that this will always fail when hitting curl functions:

```
Curl_attach_connection
Curl_llist_count
Curl_conn_is_alive
```

Which are called when using `curl_multi_socket_action` in our codebase in https://github.com/supabase/pg_net/blob/a7792bfd913c7859e14025d606624982706c2a7f/src/worker.c#L320-L336

Unfortunately this is quite hard to debug, it might be a bug in curl itself.

### DIFF
--- a/src/worker.c
+++ b/src/worker.c
@@ -232,6 +232,48 @@ static void unlock_extension(Oid ext_table_oids[static total_extension_tables]){
   UnlockRelationOid(ext_table_oids[1], AccessShareLock);
 }
 
+static void *
+net_calloc(size_t a, size_t b)
+{
+  elog(DEBUG1, "net_calloc: a(%zu), b(%zu)", a, b);
+  return palloc0(mul_size(a, b));
+}
+
+static void
+net_free(void *a)
+{
+  elog(DEBUG1, "net_free");
+  if (a)
+    pfree(a);
+}
+
+static void *
+net_malloc(size_t sz)
+{
+  elog(DEBUG1, "net_malloc: sz(%zu)", sz);
+  return sz ? palloc(sz) : NULL;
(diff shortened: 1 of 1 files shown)


steve-chavez:
> Unfortunately this is quite hard to debug, it might be a bug in curl itself.

I think this is because some curl allocations are done in sub memory contexts and the `curl_global_init_mem` is now causing some frees and mallocs to happen at the top level.

PR #229, opened 2025-08-22
fix: hardcoding the shared lib name  

steve-chavez:
Use `module_pathname` to fix this.


za-arthur (approved):
Thank you @steve-chavez, LGTM

PR #233, opened 2025-08-29
perf: don't insert on every event loop iteration

steve-chavez:
This moves `insert_curl_responses` outside of the event loop, which wasn't really needed as curl responses might not be ready on each iteration.

Moving it outside of the event loop saves cycles, and reduces CPU usage.

### DIFF
--- a/src/worker.c
+++ b/src/worker.c
@@ -335,8 +335,9 @@ void pg_net_worker(__attribute__ ((unused)) Datum main_arg) {
               );
             }
 
-            insert_curl_responses(worker_state, CurlMemContext);
           }
+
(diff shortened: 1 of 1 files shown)


src/worker.c
```diff
@@ -335,9 +335,10 @@ void pg_net_worker(__attribute__ ((unused)) Datum main_arg) {
               );
             }
 
-            insert_curl_responses(worker_state, CurlMemContext);
           }
 
+          insert_curl_responses(worker_state, CurlMemContext);
+
```
steve-chavez, line 341: Just a single line change

laurenceisla (approved):
Cool, all CPU usage % is in the single digits now for the heaviest tests!

PR #254, opened 2026-05-17
fix: flush pgstat counters from worker so autovacuum sees its writes

utkarash2991:
```
The `pg_net` background worker performs DML on `net._http_response` and `net.http_request_queue` via 
SPI but never calls `pgstat_report_stat()`. As a result, per-write counters (`n_tup_ins`, `n_tup_del`,
`n_mod_since_analyze`) for the worker's writes never reach shared stats. Autovacuum/autoanalyze
read those counters to decide when to vacuum — when they stay at zero, the launcher never schedules a
run, and `net._http_response` accumulates bloat indefinitely.
  
User-backend INSERTs into `net.http_request_queue` (via `net.http_get/post/delete`) keep that 
table's autovacuum cadence healthy in practice — user backends flush pgstat automatically via the main
loop in `tcop/postgres.c`. So the customer-visible failure is specific to `net._http_response`, which 
has no user-backend traffic to compensate. Eventually the bloated `_http_response_created_idx` 
makes the worker's expiry query (`ORDER BY created LIMIT $batch`) walk huge stretches of dead index 
entries,  yielding 20–100s DELETEs and IO-wait spikes.
```

Bug fix

Auto vacuum/analyze will not trigger on the `net._http_response`

## Fix
Call `pgstat_report_stat(false)` after each worker transaction commits. Per [`pgstat.c`](https://github.com/postgres/postgres/blob/REL_17_STABLE/src/backend/utils/activity/pgstat.c#L560-L562): 
> *"Must be called by processes that performs DML: tcop/postgres.c, logical receiver processes, SPI worker, etc. to flush pending statistics updates to shared memory."*
  
Regular user backends invoke it from the main loop after each query; background workers have no equivalent and must flush themselves.

## Tests
Two regression tests added to `test/test_worker_behavior.py`:  
1. **`test_worker_writes_increment_pgstat_counters`** — drives 30 requests through the worker and asserts `pg_stat_user_tables.n_tup_ins > 0` on `net._http_response`. Before the fix, this stays at 0; with the fix, it reflects the worker's INSERTs.
2. **`test_worker_writes_trigger_autoanalyze_on_http_response`** — end-to-end test that proves the customer-impacting symptom is resolved. Configures `autovacuum_naptime=1s` plus low per-table thresholds, drives traffic, and waits for `autoanalyze_count > 0`.

Linear: https://linear.app/supabase/issue/PSQL-1216

### DIFF
--- a/test/test_worker_behavior.py
+++ b/test/test_worker_behavior.py
@@ -442,3 +442,149 @@ def test_processing_survives_postmaster_crash():
 
     engine.dispose()
 
+
+def test_worker_writes_increment_pgstat_counters(sess, autocommit_sess):
+    """the worker's INSERTs into net._http_response must be reflected in
+    pg_stat_user_tables. Without this, autovacuum/autoanalyze can never be
+    scheduled and the table silently bloats.
+    """
+
+    # Make sure the worker is fully up before we start, otherwise we race
+    # with whatever the previous test left behind.
+    autocommit_sess.execute(text("select net.wait_until_running();"))
+
+    # Clean baseline so deltas are unambiguous.
+    autocommit_sess.execute(text(
+        "select pg_stat_reset_single_table_counters('net._http_response'::regclass);"
+    ))
+
+    # Drive a batch of requests through the worker.
+    sess.execute(text("""
+        select net.http_get('http://localhost:8080/pathological?status=200')
+        from generate_series(1,30);
+    """))
+    sess.commit()
+
+    # Wait until the worker has actually drained the queue and written all
+    # responses to net._http_response. Don't assume "30 rows" - the worker
+    # may pick up the queue in chunks depending on wake() coalescing.
+    for _ in range(20):
+        time.sleep(0.5)
+        (queue_count,) = sess.execute(text(
+            "select count(*) from net.http_request_queue;"
+        )).fetchone()
+        if queue_count == 0:
+            break
+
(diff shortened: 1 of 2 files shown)


test/test_worker_behavior.py
```diff
@@ -442,3 +442,149 @@ def test_processing_survives_postmaster_crash():
...
+    resp_ins = 0
+    resp_mod = 0
+    while time.time() < deadline:
+        (resp_ins, resp_mod) = autocommit_sess.execute(text("""
+            select n_tup_ins, n_mod_since_analyze
+            from pg_stat_user_tables where relname='_http_response';
+        """)).fetchone()
+        if resp_ins > 0:
+            break
+        time.sleep(0.5)
```
za-arthur, line 501: You can try to use `pg_stat_force_next_flush()` instead of the loop. It could make the test a little faster.

za-arthur (approved):
Good catch! LGTM. I have just a suggestion below.

test/test_worker_behavior.py
utkarash2991: I looked at this and unfortunately I think `pg_stat_force_next_flush()` won't help here — it sets a process-local flag (`pgStatForceNextFlush` is file-static in `pgstat.c`), so calling it from the test session only forces that backend's next flush. The pg_net worker is a different process, so it doesn't see the flag and still respects `PGSTAT_MIN_INTERVAL`.

PR #255, opened 2026-05-18
fix: report worker activity to pg_stat_activity

utkarash2991:
## Problem

The pg_net background worker doesn't call `pgstat_report_activity()`, so its row in `pg_stat_activity` shows up with blank `state`, `query`, `query_start`, `xact_start`, and `state_change` columns. This makes it impossible to tell from monitoring whether the worker is idle, processing a tick, or stuck in a transaction.

Regular user backends get this for free via `tcop/postgres.c`. Background workers must do it themselves.

## Fix

Add `pgstat_report_activity` transitions around the inner work loop:

- `STATE_IDLE` initially (worker starts by waiting for a wake)
- `STATE_RUNNING` when the worker enters the inner work loop
- `STATE_IDLE` when the inner loop drains and the worker goes back to waiting

`cmd_str = NULL`, matching what other background workers do (logical replication apply worker, walsender on idle). `backend_type` already identifies the row as `"pg_net <ver> worker"`, so a query-column label would be redundant.

## Verification

Sample output before/after on a local repro:

**Before** (`backend_type ilike '%pg_net%'`):

```
 pid | backend_type         | state | wait_event
-----+----------------------+-------+-----------
 218 | pg_net 0.20.2 worker |       | Extension
```

**After, at rest:**

```
 pid | backend_type         | state | wait_event
-----+----------------------+-------+-----------
 218 | pg_net 0.20.2 worker | idle  | Extension
```

**After, while processing a tick:**

```
 pid | backend_type         | state  | wait_event
-----+----------------------+--------+-----------
 218 | pg_net 0.20.2 worker | active | Extension
```

Refs PSQL-1221, follow-up to #254.

### DIFF
--- a/src/worker.c
+++ b/src/worker.c
@@ -265,6 +265,9 @@ void pg_net_worker(__attribute__((unused)) Datum main_arg) {
 
   publish_state(WS_RUNNING);
 
+  // Initial state: we go straight into the outer loop and wait for a wake.
+  pgstat_report_activity(STATE_IDLE, NULL);
+
   do {
 
     uint32 expected = 1;
@@ -273,6 +276,8 @@ void pg_net_worker(__attribute__((unused)) Datum main_arg) {
       wait_while_processing_interrupts(WORKER_WAIT_NO_TIMEOUT, &worker_should_restart);
       continue;
     }
+
+    pgstat_report_activity(STATE_RUNNING, NULL);
 
     uint64 requests_consumed = 0;
     uint64 expired_responses = 0;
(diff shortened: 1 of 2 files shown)


steve-chavez:
@utkarash2991 Can we test this somehow?

utkarash2991:
> @utkarash2991 Can we test this somehow?

@steve-chavez Added a regression test in 8ec2990. It asserts pg_stat_activity.state shows 'idle' at rest and 'active' during a slow request (using delay=2 so the active window is observable). Please let me know how this looks .

utkarash2991:
> > @utkarash2991 Can we test this somehow?
> 
> @steve-chavez Added a regression test in [8ec2990](https://github.com/supabase/pg_net/commit/8ec299029fa2543001fcebac4a0d564376425305). It asserts pg_stat_activity.state shows 'idle' at rest and 'active' during a slow request (using delay=2 so the active window is observable). Please let me know how this looks .

@steve-chavez does the changes looks good to you ? ok to merge ?

steve-chavez (approved):
LGTM 👍

PR #258, opened 2026-06-23
Port pg_net to PostgreSQL 19

devrimgunduz:
- worker.c: silence -Werror=unused-parameter for the new pg_siginfo argument PG19 expanded SIGNAL_ARGS from one parameter (postgres_signal_arg) to two (postgres_signal_arg, pg_siginfo).  The existing __attribute__((unused)) prefix only silenced the first parameter.  Add a (void) cast guarded by PG_VERSION_NUM >= 190000 in handle_sigterm and handle_sighup.

- worker.c: pass both SIGNAL_ARGS arguments to procsignal_sigusr1_handler procsignal_sigusr1_handler() now matches the new SIGNAL_ARGS signature in PG19, so the forwarding call in handle_sigusr1 must supply pg_siginfo as the second argument.

- core.c: dereference log_min_messages under PG19 In PG19, log_min_messages changed from int to int* (GUC value moved to shared memory).  Comparing a pointer to the integer constant DEBUG2 is invalid; guard with a #if block that dereferences the pointer for PG19 and keeps the direct comparison for older releases.

- core.c: adapt jsonb_headers_from_curl_handle to the revised pushJsonbValue API PG19 replaced the (JsonbParseState **, ...) -> JsonbValue* overload with a new (JsonbInState *, ...) -> void signature.  The accumulated result is now stored in JsonbInState.result (a JsonbValue*) rather than being returned by the final WJB_END_OBJECT call.  Add a PG_VERSION_NUM >= 190000 branch that:
    * declares JsonbInState on the stack (zero-initialised with = {0})
    * passes &headers (JsonbInState*) to every pushJsonbValue call
    * calls JsonbValueToJsonb(headers.result) to obtain the final Jsonb* (headers.result is already JsonbValue*, so no & is needed) The old JsonbParseState** path is preserved in the #else branch.

- event.c: replace empty () with (void) in both event_monitor definitions GCC -Werror=old-style-definition rejects functions declared with () (which in C means "unspecified arguments", a K&R-era convention) when no parameters are intended.  Changed to (void) in both the epoll and kqueue variants.

Coded by Claude, tested by me.

Per: https://github.com/pgdg-packaging/pgdg-rpms/issues/213

### DIFF
--- a/src/core.c
+++ b/src/core.c
@@ -104,7 +104,11 @@ void init_curl_handle(CurlHandle *handle, RequestQueueRow row) {
   EREPORT_CURL_SETOPT(handle->ez_handle, CURLOPT_TIMEOUT_MS, (long)handle->timeout_milliseconds);
   EREPORT_CURL_SETOPT(handle->ez_handle, CURLOPT_PRIVATE, handle);
   EREPORT_CURL_SETOPT(handle->ez_handle, CURLOPT_FOLLOWLOCATION, (long)true);
+#if PG_VERSION_NUM >= 190000
+  if (*log_min_messages <= DEBUG2) EREPORT_CURL_SETOPT(handle->ez_handle, CURLOPT_VERBOSE, 1L);
+#else
   if (log_min_messages <= DEBUG2) EREPORT_CURL_SETOPT(handle->ez_handle, CURLOPT_VERBOSE, 1L);
+#endif
 #if LIBCURL_VERSION_NUM >= 0x075500 /* libcurl 7.85.0 */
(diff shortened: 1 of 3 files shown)


steve-chavez:
@devrimgunduz Thanks! Let me get pg 19beta running on CI to prove this is working correctly :hourglass_flowing_sand:

devrimgunduz:
Hi @steve-chavez ,

Any updates? v19 beta2 is due next week.

steve-chavez:
@devrimgunduz Sorry for the late reply, opened https://github.com/supabase/pg_net/pull/266.

steve-chavez:
Solved and released on https://github.com/supabase/pg_net/releases/tag/v0.20.5. Thanks again for reporting.

Issue #227, opened 2025-08-21
Request to replace the `not null` constraint on column `created` in `net._http_response` with a check constraint and make it not inheritable


diraneyya:
# This is not a bug but a code-design proposal

I am building a scraper extension on top of `pg_net` and identified that you can do some really nifty things when you inherit from `net._http_response` and install triggers to re-route the response to other dedicated scraper buffer tables.

This is made possible with your very flexible design where the `net._http_response` has on primary key and is intentionally kept quite raw. I like this a lot.

To support my work (and others building on top of `pg_net`), I am proposing either to remove the not null constraint on the `created` column altogether or replace it with a check constraint that is not inheritable.

I would like to know how you feel about this.

diraneyya:
Immediately after posting this. I am not quite sure about it anymore (namely, that it would make a difference for me). So I would like to put it on hold and come back to it at a later point in time if needed. 

It would make sense for me to finish the functional scraper and then come back with more solid points.

diraneyya:
How it works for a simple repetitive fetch cycle:

```sql
call start_fetch_cycle('timeout', jsonb_build_object(
  'delay_ms', 1000,
  'repeat_count', 10,
  'callable_proc_name', 'foo',
  'callable_proc_args', '[]'::jsonb
));
```

`timeout` is a simple fetch-cycle type that calls an arbitrary procedure after a certain duration in milliseconds, which can be repeated for any number of times (10 times above).

The fetch in this case is to a sleep endpoint.

For more useful cycle types you can think of fetch cycles that scrape a paginated APIs, collect data in a custom buffer, and flushes them after they are merged into a target table.

Using inheritance, I allow cycle developers to inherit from `fetch_cycle_buffer`. There are two types of cycles:

1. **Simple cycles:** use the default buffer, which is immutable. The steps are deleted immediately when they are continued. This is done in an atomic swap using a before insert trigger on the default buffer table.
1. **Advanced (scrape) cycles:** use a custom buffer table that inherits from the default buffer. Using inheritance in postgres, the entries in them are still visible when querying the default buffer table. Steps in custom buffers are only archived by nullifying their `pgnet_id` field but the rows are retained until they are flushed by the cycle developer. This is necessary for scraping cycles.

I am still actively developing this.

diraneyya:
https://forgejo.orwa.tech/open-source/pg_fetch_cycle

diraneyya:
[...]

PR #259, opened 2026-07-01
fix: revoke public trigger privilege on net tables

samrose:
## Summary

  This fixes a privilege escalation issue caused by granting `ALL` table privileges on pg_net internal tables to `PUBLIC`.
 
  `ALL` includes `TRIGGER`, which allowed ordinary roles to create triggers on `net._http_response`.

## Changes

  - Replace `grant all on all tables in schema net to PUBLIC` with an explicit grant list that excludes `TRIGGER`.
  - Add an upgrade script from `0.20.3` to `0.20.4` that revokes `TRIGGER` on all tables in the `net` schema from `PUBLIC`.
  - Bump the extension version to `0.20.4`.
  - Add a regression test confirming ordinary roles cannot create triggers on `net._http_response`.

  The previous broad grant was intended to let non-`postgres` roles use pg_net after the SQL functions were changed to security invoker. That behavior should be preserved.

  The minimal safe fix is to remove only the `TRIGGER` privilege while keeping the table privileges needed for normal pg_net operation.

  ## Testing

  ```sh
  nix develop -c xpg test

  Result:

  63 passed in 92.65s (0:01:32)

  Focused regression check:

  env PYTEST_ADDOPTS='test/test_privileges.py -k test_public_roles_cannot_create_triggers_on_internal_tables' nix develop -c xpg test

  Result:

  1 passed, 3 deselected


sql/pg_net.sql
```diff
@@ -354,4 +354,4 @@ $$;
 
 grant usage on schema net to PUBLIC;
 grant all on all sequences in schema net to PUBLIC;
-grant all on all tables in schema net to PUBLIC;
+grant select, insert, update, delete, truncate, references on all tables in schema net to PUBLIC;
```
hanefi: This particular change removed more than trigger grant. For example MAINTAIN is no longer granted and users can no longer run vacuum, analyze, lock etc. This might be safe, but I'd rather see all grant privileges except trigger on this list.

Better yet, we can grant all privileges and revoke trigger immediately after that. This way it would be easier to read and maintain IMO.

This particular alternative has another important aspect: the schema will not diverge between databases that performed an upgrade from the previous version, and databases that created the extension from scratch using this migration script.

In case I am wrong about the purpose of this file, feel free to ignore this comment.

brainrake (approved):
Looks like it does what the issue describes and fixes the problem. Also does version bump same as last one.
There may be consequences I am unaware of.

sql/pg_net.sql
imor: Yep, agree with @hanefi. Let's preserve the `all` grant and just add the same line as in the upgrade sql file here as well:

```sql
revoke trigger on all tables in schema net from PUBLIC;
```

test/test_privileges.py
```diff
@@ -120,3 +120,42 @@ def test_net_on_new_role(sess):
...
+        sess.execute(text("""
+            set local role to pre_existing;
+
+            create trigger my_trigger
+            after insert on net._http_response
+            for each row
+            execute function public.my_trigger_fn();
+        """))
+
+    assert "permission denied" in str(execinfo.value)
```
imor: Test can be expanded to include the `net.http_request_queue` table as well to be really explicit that none of the tables have this privilege escalation bug. Also asserts can include the table name for explicit checks:

```suggestion
    sess.commit()

    with pytest.raises(Exception) as execinfo:
        sess.execute(text("""
            set local role to pre_existing;

            create trigger my_trigger
            after insert on net.http_request_queue
            for each row
            execute function public.my_trigger_fn();
        """))

    assert "permission denied for table http_request_queue" in str(execinfo.value)

    sess.rollback()

    with pytest.raises(Exception) as execinfo:
        sess.execute(text("""
            set local role to pre_existing;

            create trigger my_trigger
            after insert on net._http_response
            for each row
            execute function public.my_trigger_fn();
        """))

    assert "permission denied for table _http_response" in str(execinfo.value)
```

imor (approved):
@samrose left a couple of comments for improvements, otherwise LGTM. Approved to unblock you.

sql/pg_net.sql
za-arthur: Preserving `all` also would make it consistent with the upgrade script and with the state after upgrading `pg_net`

imor:
I'm taking over this bug fix from @samrose and have pushed a couple of commits. @hanefi, @za-arthur could you please do a final review before I merge it?

sql/pg_net.sql
```diff
@@ -355,3 +355,7 @@ $$;
 grant usage on schema net to PUBLIC;
 grant all on all sequences in schema net to PUBLIC;
 grant all on all tables in schema net to PUBLIC;
+
+-- we do not want users to be able to create triggers on tables in the net schema to avoid
+-- privilege escalation bugs.
+revoke trigger on all tables in schema net from PUBLIC;
```
hanefi: nit; missing endline.

hanefi (approved):

sql/pg_net.sql
imor: Fixed.

sql/pg_net.sql
imor: Fixed.

za-arthur (approved):

steve-chavez (approved):

steve-chavez (changes requested):
Just to make it clear, the vulnerability here doesn't have anything to do with SECURITY definer functions.

Regular INVOKER functions have the same problem because they're using the worker which uses the "initdb-created superuser" by default to execute the functions; but this can be changed with `pg_net.username`.

I'm not sure if this is the right fix because some users have claimed to use [TRIGGER functions on the net tables](https://github.com/supabase/pg_net/issues/227#issue-3341382791) so we would be causing a breaking change.

imor:
That's a much better alternative suggested by @steve-chavez. Basically we set `pg_net.username` to a non-superuser role like `postgres` and the worker runs the sql as that user, entirely shutting down the privilege escalation vector: now the triggers would run as the `postgres` user. Not to mention that triggers are a valid API surface we support. Since the main fix for this should be done in `supabase/postgres`:

* @steve-chavez I'm thinking of changing the scope of this PR to add a test confirming that triggers indeed run as the `pg_net.username` user.
* @samrose could you please point me to the code where `pg_net.username` GUC can be set in `supabase/postgres`, I can open a PR in that repo.

steve-chavez:
> I'm thinking of changing the scope of this PR to add a test confirming that triggers indeed run as the pg_net.username user.

Sounds good to me.

samrose:
sounds good to me to thanks @steve-chavez and @imor

PR #260, opened 2026-07-02
refactor: improve multi_socket_cb implementation

imor:
This PR improves the `multi_socket_cb` implementation for both kqueue and epoll by doing the following:

For kqueue:

* Fixes the leaks in the filters because the `sock_info->action` was not being updated. This also lead to a use-after-free bug and sometimes made the tests on macOS fail due to a segfault. All these bugs are fixed.
* Correctly checks the return value of `curl_multi_assign` and raises a Postgres error if the call fails.

For epoll:

* Correctly checks the return value of `curl_multi_assign` and raises a Postgres error if the call fails.
* Correctly handle the case when `what == CURL_POLL_INOUT` by setting the right `EPOLLIN` and `EPOLLOUT` combination in `ev.events` member.
* Sets a sentinel value in the call to `curl_multi_assign` the first time `multi_socket_cb` is called for a socket fd and removes it when `what == CURL_POLL_REMOVE`. This clarifies the intent of the `curl_multi_assign` call because the older code assignments to `socket_exists` variable were never read and were confusing.

### DIFF
--- a/src/errors.h
+++ b/src/errors.h
@@ -19,13 +19,6 @@
       ereport(ERROR, errmsg("Could not curl_multi_setopt(%s)", #opt));                             \
   } while (0)
 
-#define EREPORT_CURL_MULTI_ASSIGN(multi_handle, sockfd, sockp)                                     \
-  do {                                                                                             \
-    CURLMcode mresult = curl_multi_assign(multi_handle, sockfd, sockp);                            \
-    if (mresult != CURLM_OK)                                                                       \
-      ereport(ERROR, errmsg("Could not curl_multi_assign(%s)", curl_multi_strerror(mresult)));     \
-  } while (0)
-
 #define EREPORT_CURL_SLIST_APPEND(list, str)                                                       \
   do {                                                                                             \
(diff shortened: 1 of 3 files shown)


steve-chavez:
@imor Can we prefix this PR as `refactor:` instead of `fix:`? Two reasons:

- It doesn't need a release since Macos is not used for production. Otherwise it could be confusing for packagers like `supabase/postgres`
- Macos compatibility was only provided for development purposes.

imor:
@steve-chavez yeah I intend to clean this PR up. I added this as a draft so that I don't forget. I only needed this change to run the tests on my local setup.

steve-chavez (approved):
@imor Looks good! On second thought I'll change this to a `fix:` commit since it also changes the logic for epoll. That way the change clearly shows on the release (in case if something goes wrong).

steve-chavez:
A further improvement could be to run the loadtest on macos, that should prove pg_net works the same there

https://github.com/supabase/pg_net/blob/0ab4bae33a306316f9056fc9c87ba3f977fc6b7e/.github/workflows/main.yml#L56-L78

PR #261, opened 2026-07-03
test: add tests for privilege change when updating pg_net.username

imor:
This PR adds the following tests:

1. Test that updating the `pg_net.username` GUC from being unset to an unprivileged user continues to work after such a change.
2. Test that any triggers created on `net.http_request_queue` and `net._http_response` tables run as the same user as the background worker uses to connect.

### DIFF
--- a/test/test_privilege_migration.py
+++ b/test/test_privilege_migration.py
@@ -237,8 +237,15 @@ def test_triggers_on_net_tables_run_as_pg_net_username(sess, autocommit_sess):
         )
 
     finally:
-        sess.execute(text("drop table if exists public.trigger_witness;"))
+        # Drop the function (which cascades to the triggers on the net
+        # tables) before dropping trigger_witness. The worker's own
+        # transactions lock the net tables first and trigger_witness second
+        # (queue/response DML, then the witness trigger's INSERT); dropping
+        # trigger_witness first would acquire these same two locks in the
+        # opposite order and can deadlock with an in-flight worker
+        # transaction.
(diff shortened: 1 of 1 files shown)


test/test_privilege_migration.py
```diff
@@ -0,0 +1,254 @@
...
+# with the privileges of whoever performs DML on those tables. Since the
+# worker previously ran as the bootstrap superuser (when pg_net.username was
+# unset), that trigger code executed as superuser.
+#
+# The fix is to configure pg_net.username to a non-superuser role. This
+# suite checks that projects which already have data in the net tables
+# (queued requests, historical responses) keep working after making that
+# change - i.e. that Postgres' table-level GRANTs (not row "ownership")
+# govern what the worker can do, so switching users doesn't orphan
+# pre-existing rows.
```
steve-chavez, line 18: I understand the motivation here, but I see some drawbacks:

- The tests are long and look hard to maintain, it's a lot of steps/text
- We're testing postgres privilege behavior intertwined with pg_net calls, which makes it all more verbose

IMHO for extra assurance and lower maintenance, the tests could be added on `supabase/postgres` and be: a db object created with `supabase_admin` on public, can it still be read/written by the `postgres` role? 

But TBH, `postgres` is quasi-superuser on `supabase/postgres` and if something broke with the pg_net.username migration likely the user was already doing something very wrong (🥷 ) or we're missing some edge case. In any case I don't think we can come up with a good test, so I'd be fine going forward without it.

imor: Okay, I'll move these to `supabase/postgres`.

PR #262, opened 2026-07-03
build: could not build with pg versions older than 16 on macos

imor:
This PR fixes the build on macos for Postgres versions older than 16. Without this fix, when trying to run tests on such a version the build failed with this message:

```
$ xpg -v 12 test
make: *** No rule to make target 'pg_net.dylib', needed by 'build-12/pg_net.dylib'.  Stop.
```


steve-chavez (approved):
LGTM 👍

PR #263, opened 2026-07-03
test: fix flaky tests by replacing hardcoded sleeps with polls

imor:
This PR started off as an attempt to fix flaky tests but has grown into a mini overhaul of the tests. The changes include:

* Fixed many causes of flakiness, including hardcoded `time.sleep(...)` calls and race conditions and assumptions in the test code. The improvement is evidenced in the last 12 or so commits in this PR. Before this PR, the tests would fail quite often. Examples: [one](https://github.com/supabase/pg_net/actions/runs/30274356310), [two](https://github.com/supabase/pg_net/actions/runs/28860068242), [three](https://github.com/supabase/pg_net/actions/runs/28860065869), [four](https://github.com/supabase/pg_net/actions/runs/28754378777), and [many](https://github.com/supabase/pg_net/actions/runs/28721006742) [more](https://github.com/supabase/pg_net/actions/runs/28720661680). The new code waits for certain events to happen instead of hardcoded timeouts which is more robust.
* Made the code more readable by factoring out common code into `common.py` which also makes test smaller and easier to read.
* Removed some commented out code.
* Moved cleanup logic in tests into finally blocks to ensure one failing tests doesn't make other tests fail, improving test isolation.
* Formatted the code according to python standards.
* Added support for running a subset of tests locally during development. Updated `CONTRIBUTING.md` to show how to run a single or a subset of tests.
* Made sure on pushing a commit to a PR branch only one set of tests run. Previously two sets of tests would run in this case. Merging into `master` still triggers a separate test run.

Further improvement planned in a separate PR: currently test on each PG version unnecessarily pulls in all PG versions in nix store via xpg. This is ~700 MB worth of data pulled from the cache. This can be reduced if a single test could only copy that specific version of PG.

### DIFF
--- a/test/common.py
+++ b/test/common.py
@@ -1,0 +1,329 @@
+import time
+from sqlalchemy import create_engine, text
+from sqlalchemy.orm import Session
+
+
+def http_request(sess, query):
+    """
+    Execute query and commit to wake up the background worker.
+
+    The query should return a single row with a single column
+    containing the request id of the request. Returns the request
+    id
+    """
+    request_id = sess.execute(query).scalar_one()
+    # Commit to wakeup background worker
+    sess.commit()
+    return request_id
+
+
+def http_requests(sess, query):
+    """
+    Execute query and commit to wake up the background worker.
+
+    The query usually contains multiple http requests. Returns
+    the reqeust id of the first request in the query
+    """
+    (request_id,) = sess.execute(query).first()
+    # Commit to wakeup background worker
+    sess.commit()
+    return request_id
+
+
+def collect_response_sync(sess, request_id):
+    """
+    Wait for request with request_id to complete and return its response.
+
+    Flattens net._http_collect_response's nested composite return type
+    (status, message, response(status_code, headers, body)) into a single
+    row, so callers get a name-addressable mapping (e.g. result["body"])
+    instead of indexing into a stringified nested tuple.
+    """
+    return sess.execute(
+        text(
+            """
+        select
+            status,
+            message,
+            (response).status_code,
+            (response).headers,
+            (response).body
+        from net._http_collect_response(:request_id, async:=false);
+    """
+        ),
+        {"request_id": request_id},
+    ).mappings().fetchone()
+
+
+def is_worker_up(autocommit_sess):
+    """
+    Returns a function that checks whether worker is up or not
+
+    The returned function captures autocommit_sess argument and
+    uses it to track worker status. 
+    """
+
+    def fetch():
+        (worker_is_up,) = autocommit_sess.execute(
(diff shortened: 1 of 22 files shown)


test/common.py
```diff
@@ -0,0 +1,329 @@
...
+
+    But it has a race condition in which this function might
+    return before the worker has restarted. This happens because
+    net.worker_restart() returns immediately after setting a flag
+    to indicate to the core worker loop to restart. Then the
+    net.wait_until_running() function waits for the worker state to
+    become WS_RUNNING. But it can read the state from either the worker
+    before the restart or after. In the first case it returns before
+    the worker has restarted properly, and in the second case it
+    behaves correctly.
```
steve-chavez, line 307: Right, overall the condition variables in the extension had the goal to remove sleeps from the pytests too but those are not yet stable. 

https://github.com/supabase/pg_net/blob/c97a92aea57965fc4fcef84e8570f6cdc02bec21/src/worker.c#L65-L74

steve-chavez (approved):
Overall LGTM.

> Formatted the code according to python standards.

Note: to make reviewing easier is better to do linting on a separate PR. It should also be enforced on CI.

steve-chavez:
The commits here are not looking that good for git history -- critical for when a bug arises on prod.

Also https://github.com/supabase/pg_net/pull/263/changes/a64f20d5d59771834d8feb5468f848dcfe037afd is labeled as `refactor:` but it really is a `test:` (because it doesn't change core logic).

It looks to me all of these can be squashed onto 3 commits:

- `test: ...`
- `docs: ...`
- `ci: ...`

@imor Before merging, could you do that? :pray: 

**Edit**: Non-blocking request, but would be good to have. Feel free to merge if you disagree.

imor:
> The commits here are not looking that good for git history -- critical for when a bug arises on prod.
> 
> Also [a64f20d](https://github.com/supabase/pg_net/commit/a64f20d5d59771834d8feb5468f848dcfe037afd) is labeled as `refactor:` but it really is a `test:` (because it doesn't change core logic).
> 
> It looks to me all of these can be squashed onto 3 commits:
> 
> * `test: ...`
> * `docs: ...`
> * `ci: ...`
> 
> @imor Before merging, could you do that? 🙏
> 
> **Edit**: Non-blocking request, but would be good to have. Feel free to merge if you disagree.

No worries, will rewrite into a cleaner history @steve-chavez. I know this was not the easiest of PRs to review so thanks for you time 🙏🏼

test/common.py
imor: TIL, thanks for this comment. Will try to improve it in a future PR, resolving the comment for now.

PR #271, opened 2026-07-31
tests: do not download all PG versions

imor:
xpg version 2.5.0 added support for downloading only one version of Postgres. This PR bumps xpg to 2.5.0 and uses it to download only the versions needed in a test workflow. This change downloads ~164 MBs less than previous versions.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -43,16 +43,16 @@
         "nixpkgs": "nixpkgs_2"
       },
       "locked": {
-        "lastModified": 1785519554,
+        "lastModified": 1785529904,
         "narHash": "sha256-DDIeJCHRtmvB7yocWTo1MISIGf0DsUmyLw6ISuEV4e0=",
-        "owner": "imor",
+        "owner": "steve-chavez",
         "repo": "xpg",
-        "rev": "c59ee9337c0b036a0be3ff4184626c344a818f4a",
+        "rev": "e7d47eb58f19168b5facc6b78a64f6cc7f04bbcb",
         "type": "github"
       },
       "original": {
         "owner": "steve-chavez",
-        "ref": "v2.4.0",
+        "ref": "v2.5.0",
         "repo": "xpg",
         "type": "github"
       }
--- a/flake.lock
+++ b/flake.lock
@@ -43,11 +43,11 @@
         "nixpkgs": "nixpkgs_2"
       },
       "locked": {
-        "lastModified": 1782266378,
-        "narHash": "sha256-moRjdGvymGEyREU4++qpNzDuUN+nVC69CUYjhuY4Yzs=",
-        "owner": "steve-chavez",
+        "lastModified": 1785519554,
+        "narHash": "sha256-DDIeJCHRtmvB7yocWTo1MISIGf0DsUmyLw6ISuEV4e0=",
+        "owner": "imor",
         "repo": "xpg",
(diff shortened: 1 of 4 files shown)


steve-chavez:
@imor Why is this needed? I'm seeing the latest CI time to be around 2 mins https://github.com/supabase/pg_net/commit/f8a061c2bacd5ce80334ffff10c2d21cca61cb80

imor:
> @imor Why is this needed? I'm seeing the latest CI time to be around 2 mins [f8a061c](https://github.com/supabase/pg_net/commit/f8a061c2bacd5ce80334ffff10c2d21cca61cb80)

Thinking was that downloading less data from cache will improve build times but there's no material difference even though we are downloading a smaller size with this PR's change (shaving off ~164 MB):

Old:

```
these 515 paths will be fetched (735.26 MiB download, 3858.34 MiB unpacked):
```

New:
```
these 486 paths will be fetched (571.08 MiB download, 3337.61 MiB unpacked):
```

I guess network download is not on the critical path.

steve-chavez:
> but there's no material difference even though we are downloading a smaller size with this PR's change (shaving off ~164 MB):
> I guess network download is not on the critical path.

Right, I guess it can't hurt though so I'll proceed with the review.

flake.lock
```diff
@@ -43,11 +43,11 @@
...
       "locked": {
-        "lastModified": 1782266378,
-        "narHash": "sha256-moRjdGvymGEyREU4++qpNzDuUN+nVC69CUYjhuY4Yzs=",
-        "owner": "steve-chavez",
+        "lastModified": 1785519554,
+        "narHash": "sha256-DDIeJCHRtmvB7yocWTo1MISIGf0DsUmyLw6ISuEV4e0=",
+        "owner": "imor",
         "repo": "xpg",
-        "rev": "d81725666a193c644fe0db96a988ece66edf460c",
+        "rev": "c59ee9337c0b036a0be3ff4184626c344a818f4a",
```
steve-chavez: Can be updated here now that https://github.com/steve-chavez/xpg/releases/tag/v2.5.0 is released:

https://github.com/supabase/pg_net/blob/a8299b11182ea5c974f5e89ae83e70e9e44e9e8f/flake.nix#L16-L18

imor: Done @steve-chavez. flake now points to the released 2.5.0 version.

steve-chavez (approved):

steve-chavez:
Github stacks are new to me but this looks good to merge independently of the other PR

imor:
> Github stacks are new to me but this looks good to merge independently of the other PR

I'm in the same boat, was experimenting with them and the basic idea is that you can keep interdependent PR branches updated automatically with GH stacks instead of manually rebasing when a dependency branch has new commits. I just wanted this PR to reuse the flaky test fixes, but as you said it's not really dependent on the other PR, so I'll detach it from the stack and merge.

imor:
> > but there's no material difference even though we are downloading a smaller size with this PR's change (shaving off ~164 MB):
> > I guess network download is not on the critical path.
> 
> Right, I guess it can't hurt though so I'll proceed with the review.

By the way, the slowest test is `test-on-macos` which also now downloads less:

Old:

```
these 472 paths will be fetched (912.08 MiB download, 4747.92 MiB unpacked):
```

New:

```
these 438 paths will be fetched (773.32 MiB download, 4251.33 MiB unpacked):
```

PR #275, opened 2026-08-16
fix: reject request headers containing CR or LF

### LINKED ISSUE #274: pg_net doesn't reject headers with \r or \n in them
# Bug report

- [x] I confirm this is a bug with Supabase, not with my own application. 
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

pg_net doesn't reject headers with \r or \n in them leading to downstream http parsers getting confused about where the headers end and body begins. In one reported instance a header value with a trailing \n prematurely ended the header part because libcurl also appends \r\n at the end of headers (see https://curl.se/libcurl/c/CURLOPT_HTTPHEADER.html)

## To Reproduce

Add a header value with \n at the end and see the downstream http parser will see any header after the poisoned header a part of the body. Supabase's edge functions also parse the request like this.

## Expected behavior

Request with \r or \n in headers should be rejected when it's being made and should never reach the downstream http server.

hamodywe:
Closes #274.

Header values went straight to `curl_slist_append` with no validation. libcurl appends the CRLF terminator itself, so a value carrying its own CR or LF closes the header early and everything after it is read by the receiving parser as further headers or as the start of the body. A value interpolated from untrusted input can therefore inject headers into the outgoing request, or split it.

This rejects any header containing CR or LF before it reaches libcurl, so a poisoned request fails instead of being sent. The check is in `pg_text_array_to_slist` rather than in the SQL API so it also covers rows inserted directly into `net.http_request_queue`, and it uses the same `ereport(ERROR, ...)` style already used in that function.

Only the header name is included in the error. The value is the attacker-controlled part and commonly holds credentials, so it must not be copied into the server log; when the CR or LF appears before any `:`, the reported name is truncated at that point so the smuggled text cannot reach the log that way either.

**Verification**

- Builds clean in-tree against PostgreSQL 17 (`postgresql-server-dev-17`, `libcurl4-openssl-dev`), including the repo's `-Werror -Wextra -Wall` settings.
- The matching logic was exercised standalone over 11 cases: ordinary headers (including `X-Empty:` and a header with no colon) are unaffected; LF, CR, CRLF, trailing-LF and leading-LF are all rejected; and the no-colon-before-CRLF case confirms the injected text never appears in the error.
- `test/test_http_header_crlf.py` adds regression tests in the style of `test_http_malformed_headers.py`, including one asserting ordinary headers are still sent. I was not able to run the Nix `net-with-nginx` harness locally, so those tests are unrun on my side — worth a look in CI.

### DIFF
--- a/test/test_http_header_crlf.py
+++ b/test/test_http_header_crlf.py
@@ -1,0 +1,77 @@
+from sqlalchemy import text
+from common import collect_response_sync, http_request
+
+
+def test_http_request_header_with_lf_is_rejected(sess):
+    """
+    A header value ending in a line feed must not be sent.
+
+    libcurl appends the CRLF terminator itself, so an embedded LF closes the
+    header early and the receiving parser reads whatever follows as more
+    headers or as the body (request splitting).
+    """
+
+    request_id = http_request(
+        sess,
+        text(
+            r"""
+        select net.http_get(
+            url:='http://localhost:8080/headers',
+            headers:=jsonb_build_object('Injected', 'value' || chr(10) || 'X-Smuggled: 1')
+        );
+    """
+        ),
+    )
+
+    response = collect_response_sync(sess, request_id)
+
+    assert response is not None
+    assert response["status"] == "ERROR"
+    assert "carriage return or line feed" in response["message"]
+    # The header name is safe to surface; the value may hold credentials.
+    assert "Injected" in response["message"]
+    assert "X-Smuggled" not in response["message"]
+
+
+def test_http_request_header_with_cr_is_rejected(sess):
+    """A carriage return is rejected for the same reason as a line feed."""
+
+    request_id = http_request(
+        sess,
+        text(
+            r"""
+        select net.http_get(
+            url:='http://localhost:8080/headers',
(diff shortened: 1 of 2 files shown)


test/test_http_header_crlf.py
```diff
@@ -0,0 +1,77 @@
...
+        );
+    """
+        ),
+    )
+
+    response = collect_response_sync(sess, request_id)
+
+    assert response is not None
+    assert response["status"] == "ERROR"
+    assert "carriage return or line feed" in response["message"]
```
imor, line 30: Can the assert compare the full error message instead of just an `in` comparison. That will clearly show the exact error message that a user will see.

imor:
Thanks for the PR @hamodywe . Could you please:

1. Fix the failing CI checks.
2. Expand the test coverage a bit by testing that CR/LF is rejected if it: a) is both before and the `:`. Basically exercise all code paths in the `reject_crlf_in_header` function.

PR #278, opened 2026-09-07
fix: minor improvements (xact callback leak, `pg_net.ttl` validation)

### LINKED ISSUE #268: Validate `ttl` config
# Chore

Validate custom `pg_net.ttl` configs

## Describe the chore

The `pg_net.ttl` config is an interval type, but it doesn't perform any validation checks in the [code](https://github.com/supabase/pg_net/blob/a8299b11182ea5c974f5e89ae83e70e9e44e9e8f/src/worker.c#L504):
```c
DefineCustomStringVariable("pg_net.database_name", "Database where the worker will connect to",
                             NULL, &guc_database_name, "postgres", PGC_SU_BACKEND, 0, NULL, NULL,
                             NULL);
```

Based on the `DefineCustomStringVariable` [definition](https://github.com/postgres/postgres/blob/bfe5c4b/src/include/utils/guc.h#L393-L402):

```c
extern void DefineCustomStringVariable(const char *name,
									   const char *short_desc,
									   const char *long_desc,
									   char **valueAddr,
									   const char *bootValue,
									   GucContext context,
									   int flags,
									   GucStringCheckHook check_hook,
									   GucStringAssignHook assign_hook,
									   GucShowHook show_hook) pg_attribute_nonnull(1, 4);
```

validation likely would have to be added to the `GucStringCheckHook` parameter. [cybrosys](https://www.cybrosys.com/research-and-development/postgres/how-to-understand-and-implement-guc-hooks-in-postgresql) has an article on the topic. Clickhouse also has a [good article](https://clickhouse.com/blog/taming-postgres-guc-extra-data) that may be worth referencing.

The function [interval_in](https://github.com/postgres/postgres/blob/d68576bcfc7227d94c4391a3c34aadb8dbf8a3c7/src/backend/utils/adt/timestamp.c#L890) can be called against the input string to validate that it is in interval format.

## Additional context

Tried to update the MGMT-API/CLI to expose the `ttl` and `batch_size` settings:
- https://linear.app/supabase/issue/CE-3/add-pg-net-settings-to-cli

Came across an issue related to type interval validation. It can be circumvented with API level checks, but I think it would be better if the extension also prevented invalid entries. During testing, I found it to be trivial to break the background worker by running something like:

```sql
ALTER SYSTEM SET pg_net.ttl = '1 mins';
SELECT net.worker_restart();
```

utkarash2991:
Minor improvements in pg_net:

- Register the `wake_at_commit` transaction callback once per backend instead of once per transaction. Fixes a per-backend leak in `TopMemoryContext` and ever-slower `CallXactCallbacks` on long-lived connections. Measured 200,120 bytes of growth over 5000 transactions before, 0 after.
- Validate `pg_net.ttl` when it is set: invalid and negative intervals are rejected at `ALTER SYSTEM` time with the interval parser's message, instead of being accepted and failing inside the worker. Closes #268.

Tested on PG 13, 17 (full suite) and 19.

Tracking: [PSQL-1652](https://linear.app/supabase/issue/PSQL-1652)

### DIFF
--- a/test/test_ttl_config.py
+++ b/test/test_ttl_config.py
@@ -6,6 +6,19 @@ from common import collect_response_sync, http_request, restart_worker
 from common import wait_for_response_count, wakeup_worker
 
 
+def assert_pg_error(excinfo, expected_message, expected_detail):
+    """
+    Assert the exact primary message and DETAIL line of the PostgreSQL error
+    wrapped in a SQLAlchemy DBAPIError, i.e. the full error a user would see:
+
+        ERROR:  <expected_message>
+        DETAIL:  <expected_detail>
+    """
+    diag = excinfo.value.orig.diag
+    assert diag.message_primary == expected_message
+    assert diag.message_detail == expected_detail
+
+
 def reset_ttl(autocommit_sess):
     autocommit_sess.execute(text("alter system reset pg_net.ttl"))
     autocommit_sess.execute(text("select pg_reload_conf()"))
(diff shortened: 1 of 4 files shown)


test/test_ttl_config.py
```diff
@@ -0,0 +1,91 @@
...
+    An invalid interval must be rejected at ALTER SYSTEM time
+    instead of making the worker fail when it tries to delete
+    expired responses (issue #268)
+    """
+    for bad in ["1 blah", "yesterday", "abc", "1 minute 10 potatoes"]:
+        with pytest.raises(DBAPIError) as excinfo:
+            autocommit_sess.execute(text(f"alter system set pg_net.ttl to '{bad}'"))
+        msg = str(excinfo.value)
+        assert 'invalid value for parameter "pg_net.ttl"' in msg
+        assert "invalid input syntax for type interval" in msg
```
imor: Can we assert the value of the full error message, that will make it easy to see the complete error we expect users to see as well.

imor (approved):
Test asserts can be improved a bit, otherwise looks good.

PR #279, opened 2026-09-08
docs: fix README timeout defaults and add http_delete body param

bookingminds:

Docs update. The README's published function signatures do not match the shipped functions.

The README documents timeout_milliseconds int default 1000 for net.http_get and net.http_post, and default 2000 for net.http_delete. It also shows net.http_delete without a body parameter.

## What is the new behavior?

All three signatures show timeout_milliseconds int default 5000, and net.http_delete includes its body parameter, matching sql/pg_net.sql on master:

create or replace function net.http_delete(
    url text,
    params jsonb default '{}'::jsonb,
    headers jsonb default '{}'::jsonb,
    timeout_milliseconds int default 5000,
    body jsonb default null
)

## Additional context

The same two errors, with a different wrong timeout value, appear in the Supabase docs guide for pg_net. I opened a separate PR for that page.

### DIFF
--- a/README.md
+++ b/README.md
@@ -340,6 +340,7 @@ net.http_delete(
     parallel safe
     language plpgsql
     security definer
+```
 
 ### Examples:
 The following examples use the [Dummy Rest API](https://dummy.restapiexample.com/employees).
--- a/README.md
+++ b/README.md
@@ -190,7 +190,7 @@ net.http_get(
     -- key/values to be included in request headers
     headers jsonb default '{}'::jsonb,
     -- the maximum number of milliseconds the request may take before being cancelled
-    timeout_milliseconds int default 1000
+    timeout_milliseconds int default 5000
 )
     -- request_id reference
     returns bigint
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -338,7 +340,6 @@ net.http_delete(
     parallel safe
     language plpgsql
     security definer
-```
```
utkarash2991, line 341: Overall the Pr looks good to me. With a nit comment: I think we mistakenly removed line 341 i.e ``` . Can you revert that line

utkarash2991:
hey @bookingminds I took the liberty of fixing the minor comment on the Pr. Hope thats ok .  Rest all looks good to me. Going ahead with the merge.

utkarash2991 (approved):

bookingminds:
Thanks for catching that and for fixing it yourself rather than bouncing
it back  the closing fence was collateral from editing in the GitHub web
editor, and I should have re-read the rendered diff before pushing. I clone
and diff locally now. Appreciated.

PR #281, opened 2026-09-10
fix: read response headers from the final redirect hop

utkarash2991:
When a request is redirected, the `headers` stored in `net._http_response` were taken from the first response (the redirect) while `status_code`, `content` and `content_type` came from the final one. This reads the headers from the final response as well.

Adds a test that follows the existing `/redirect_me` -> `/to_here` redirect in the nginx harness and checks the stored headers belong to the final response.

### DIFF
--- a/src/core.c
+++ b/src/core.c
@@ -221,7 +221,7 @@ static Jsonb *jsonb_headers_from_curl_handle(CURL *ez_handle) {
   PG_JSONB_INIT_STATE(headers);
   (void)PG_JSONB_PUSH(headers, WJB_BEGIN_OBJECT, NULL);
 
-  while ((header = curl_easy_nextheader(ez_handle, CURLH_HEADER, 0, prev))) {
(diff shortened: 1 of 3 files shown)


imor (approved):
LGTM!

src/core.c
```diff
@@ -221,7 +221,7 @@ static Jsonb *jsonb_headers_from_curl_handle(CURL *ez_handle) {
   PG_JSONB_INIT_STATE(headers);
   (void)PG_JSONB_PUSH(headers, WJB_BEGIN_OBJECT, NULL);
 
-  while ((header = curl_easy_nextheader(ez_handle, CURLH_HEADER, 0, prev))) {
+  while ((header = curl_easy_nextheader(ez_handle, CURLH_HEADER, -1, prev))) {
```
steve-chavez, line 224: It's not obvious what this number represents, so a variable or macro would've been nice here.

utkarash2991: Sure @steve-chavez that makes sense to me. Going to cover this in my next Pr.

PR #285, opened 2026-09-15
fix: bound request timeouts

utkarash2991:
We pass `timeout_milliseconds` straight to `CURLOPT_TIMEOUT_MS` and libcurl treats 0 as no timeout. So a request with 0 timeout to a server which accepts the connection but never responds blocks the whole batch loop and the worker never comes out of it. `net.worker_restart()` also doesn't help here as the restart flag is only checked between batches.

What this PR does:

- Adds a GUC `pg_net.max_timeout_ms` (default 600000, `PGC_SUSET`).
- The worker rejects any request whose timeout is outside `1..pg_net.max_timeout_ms`: it is not sent and gets an `ERROR` response with the message `timeout_milliseconds must be between 1 and <max> (pg_net.max_timeout_ms), got <value>`. 

Behaviour change: a 0 or negative or oversized timeout used to hang the worker. Now the request fails with an error response and the batch carries on. Everything in range is untouched.


test/test_http_timeout.py
```diff
@@ -149,3 +150,45 @@ def test_many_slow_mixed_with_fast(sess):
...
+
+    with pytest.raises(Exception) as execinfo:
+        sess.execute(text(
+            """
+            insert into net.http_request_queue(method, url, timeout_milliseconds)
+            values ('GET', 'http://localhost:8080/pathological?status=200', 0);
+        """
+        ))
+
+    assert 'http_request_queue_timeout_milliseconds_check' in str(execinfo.value)
```
imor: Parameterize this test too:


```suggestion
@pytest.mark.parametrize("timeout", [0, -1, 600001])
def test_direct_insert_rejects_timeout_out_of_range(sess, timeout):
    """The bound is enforced on the queue table, not just the request functions"""

    with pytest.raises(Exception) as execinfo:
        sess.execute(text(
            f"""
            insert into net.http_request_queue(method, url, timeout_milliseconds)
            values ('GET', 'http://localhost:8080/pathological?status=200', {timeout});
        """
        ))

    assert 'http_request_queue_timeout_milliseconds_check' in str(execinfo.value)
```

imor: Also compare complete error message instead of just asserting contains.

test/test_http_timeout.py
```diff
@@ -149,3 +150,45 @@ def test_many_slow_mixed_with_fast(sess):
...
+        sess.execute(text(
+            f"""
+            select net.http_get(url := 'http://localhost:8080/pathological?status=200', timeout_milliseconds := {timeout});
+        """
+        ))
+
+    assert 'http_request_queue_timeout_milliseconds_check' in str(execinfo.value)
+
+
+def test_http_get_accepts_max_timeout(sess):
```
imor: Add corresponding min timeout test too?

utkarash2991: Taken care in latest revision.

test/test_http_timeout.py
```diff
@@ -149,3 +150,81 @@ def test_many_slow_mixed_with_fast(sess):
...
+            f"""
+            insert into net.http_request_queue(method, url, timeout_milliseconds)
+            values ('GET', 'http://localhost:8080/pathological?status=200', {timeout});
+        """
+        ))
+
+    assert execinfo.value.orig.diag.message_primary == CHECK_VIOLATION
+
+
+def test_worker_clamps_timeout_of_rows_queued_before_the_bound(sess):
```
imor: In future we could also think of having CI tests in which we run upgrades and assert on properties we want upgrades to uphold.

imor (approved):

test/test_http_timeout.py
```diff
@@ -149,3 +150,81 @@ def test_many_slow_mixed_with_fast(sess):
...
+    """
+    ))
+
+    response = collect_response_sync(sess, request_id)
+
+    assert response is not None
+
+
+@pytest.mark.parametrize("timeout", [0, -1, 600001])
+def test_direct_insert_rejects_timeout_out_of_range(sess, timeout):
```
AndrewJackson2020: Is it necessary to test inserting into table and using the convenience function? Seems to me like they test the same code paths and we can just test the function to test the entire execution.

test/test_http_timeout.py
```diff
@@ -149,3 +150,81 @@ def test_many_slow_mixed_with_fast(sess):
...
+            insert into net.http_request_queue(method, url, timeout_milliseconds)
+            values ('GET', 'http://localhost:8080/pathological?status=200', {timeout});
+        """
+        ))
+
+    assert execinfo.value.orig.diag.message_primary == CHECK_VIOLATION
+
+
+def test_worker_clamps_timeout_of_rows_queued_before_the_bound(sess):
+    """Rows that predate the check constraint are still sent, with their timeout clamped"""
```
AndrewJackson2020: Can we add some language here about what version transition we are testing?

test/test_http_timeout.py
AndrewJackson2020: I agree with this. I would much rather the tests do something like:
1. install older version of extension
2. setup (insert rows, whatever)
3. upgrade to latest or target version
4. run tests

as opposed to cherry pick aspects of the migration script in the pytest code.

src/core.c
```diff
@@ -44,12 +44,20 @@ static struct curl_slist *pg_text_array_to_slist(ArrayType *array, struct curl_s
...
+  PG_RETURN_INT32(MAX_TIMEOUT_MS);
+}
+
 void init_curl_handle(CurlHandle *handle, RequestQueueRow row) {
   handle->id        = row.id;
   handle->body      = makeStringInfo();
   handle->ez_handle = curl_easy_init();
 
   handle->timeout_milliseconds = row.timeout_milliseconds;
+  if (handle->timeout_milliseconds <= 0 || handle->timeout_milliseconds > MAX_TIMEOUT_MS) {
```
AndrewJackson2020: I'm a bit confused about this. The upgrade scripts seems to make this impossible via the below statement, does it not? Personally I think I would prefer to handle this at the SQL level. Are we concerned about a mismatch between the c .so and the customer not upgrading the in database extension? IMO we should try to not expose the customer to breaking changes without them running upgrade on the extension.

```
update net.http_request_queue
  set timeout_milliseconds = net._max_timeout_ms()
  where timeout_milliseconds <= 0 or timeout_milliseconds > net._max_timeout_ms();
```

utkarash2991: @AndrewJackson2020 we need this change as because the binary and the extension SQL don't move together. The `.so` is replaced whenever the image or package is upgraded but the constraint only appears when we run `ALTER EXTENSION UPDATE` . So there's a window where the constraint doesn't exist yet and the new worker still has to deal with out-of-range rows.
It also matters for the hung projects this PR is about. After a worker restart the 0-timeout row is back in the queue and the worker picks it up again, so the upgrade script can't even take its lock until that batch finishes. With this line the batch finishes, and then the upgrade can run.

Regarding the breaking changes, the condition here just tries to ensure that the batch finishes, which otherwise can go to hung state forever.

test/test_http_timeout.py
utkarash2991: YEs they are testing the same code path. I just added it for sake of completeness purpose.

test/test_http_timeout.py
utkarash2991: Added in the latest revision

test/test_http_timeout.py
AndrewJackson2020: IMO if they test the same code path it's not buying us anything other than slower testing times, especially since we can't run these tests in parallel (or at least its not documented).

src/core.c
AndrewJackson2020: That's fine. I did not realize upgrading the extension was not possible under those circumstances. I would just put a comment in. This logic presumably can and should be removed in the future (after we have migrated instances off of the old version).

sql/pg_net.sql
```diff
@@ -7,6 +7,13 @@ check (
   or value ilike 'delete'
 );
 
+-- API: Private
+create or replace function net._max_timeout_ms()
+    returns int
+    language 'c'
+    immutable
+as 'MODULE_PATHNAME';
```
steve-chavez: It's a really weird pattern to have this as function when it should be a GUC. If it's a function, dropping it is possible and leaves `pg_net` in a broken state, with a GUC there's no such possibility. 

This also gives us the chance to make it configurable, if we give it `PGC_SUSET` then the managed platform can decide whether to give that capability to end users or not.

src/core.c
steve-chavez: > The .so is replaced whenever the image or package is upgraded but the constraint only appears when we run ALTER EXTENSION UPDATE

The above would be a non-issue if we go ahead with a GUC as [mentioned above](https://github.com/supabase/pg_net/pull/285/changes#r4021507748).

utkarash2991: Added a GUC in the latest commit.

AndrewJackson2020:
@utkarash2991 can you make sure this is rebased so that it is picking up the new CI checks?

sql/pg_net.sql
```diff
@@ -126,6 +126,11 @@ declare
     request_id bigint;
     params_array text[];
 begin
+    if timeout_milliseconds < 1 or timeout_milliseconds > current_setting('pg_net.max_timeout_ms')::int then
+        raise exception 'timeout_milliseconds must be between 1 and %', current_setting('pg_net.max_timeout_ms')
+            using errcode = 'invalid_parameter_value';
+    end if;
```
steve-chavez: Q: Why don't we make this validation on the C code? It looks important enough to have it there.

That would save us a SQL migration, deduplicates code, plus leaves no chance a user changes the SQL function and causes worker malfunction. (this is even more important considering self-hosters)

test/test_http_timeout.py
```diff
@@ -149,3 +150,79 @@ def test_many_slow_mixed_with_fast(sess):
...
+
+OUT_OF_RANGE = "timeout_milliseconds must be between 1 and 600000"
+
+
+@pytest.mark.parametrize("timeout", [0, -1, 600001])
+def test_http_get_rejects_timeout_out_of_range(sess, timeout):
+    """Requests must have a finite lifetime, so zero, negative and oversized timeouts are rejected"""
+
+    with pytest.raises(Exception) as execinfo:
+        sess.execute(text(
```
AndrewJackson2020: Can we update these tests to use psycopg directly instead of sqlalchemy?

AndrewJackson2020: I added some infra for this here.

https://github.com/supabase/pg_net/commit/5d0606486cab7478e6c8565988e25c38e297d95b

sql/pg_net.sql
utkarash2991: @steve-chavez initially I did it in SQL so that requests with bad timeouts never make it into the queue.
Moved it entirely to C now. The worker clamp is the only mechanism: out of range timeouts are bounded to pg_net.max_timeout_ms when the request is picked up, no SQL changes and no migration. The GUC is read by the worker, so a reload applies it.

test/test_http_timeout.py
utkarash2991: Done in the latest commit.

AndrewJackson2020: Ty

sql/pg_net.sql
steve-chavez: @utkarash2991 Awesome! Just look at how much smaller the PR got :rocket:

test/test_http_timeout.py
```diff
@@ -162,3 +164,59 @@ def test_many_slow_mixed_with_fast(sess):
...
+
+    request_id = pg_http_request(
+        conn,
+        "select net.http_get(url := 'http://localhost:8080/pathological?status=200', timeout_milliseconds := %s)",
+        (timeout,),
+    )
+
+    response = pg_collect_response(conn, request_id)
+
+    assert response["status"] == "SUCCESS"
```
steve-chavez: I think we should refine this. Two options:

- Just fail when the timeout surpasses the GUC one. IMO the best.
- Silently clamp it as it's done now, but log a warning.

utkarash2991: @steve-chavez went through the first approach. Does it looks good now?

steve-chavez: @utkarash2991 Looks good. I guess it's strictly a breaking change, but a good thing since otherwise users can cause worker malfunction.

steve-chavez (approved):

PR #286, opened 2026-09-15
upgrade tests to use psycopg3

AndrewJackson2020:
We use sqlalchemy for as our ORM, the default driver for postgres is psycopg2. Psycopg3 exists. Psycopg2 is in a maintenance only mode where it only receives bug fixes and only if its not too much pain to fix them. Let move to psycopg3.


test/test_privileges.py
```diff
@@ -59,10 +59,11 @@ def test_net_on_new_role(sess):
     sess.execute(text("""
         create role another;
     """))
+    sess.execute(text("set local role to another;"))
 
     (request_id, current_user) = sess.execute(text(
         """
-        set local role to another;
+
```
hanefi:
```suggestion
```

hanefi (approved):
LGTM

PR #287, opened 2026-09-15
start using psycopg directly in tests

AndrewJackson2020:
Currently we use sqlalchemy in the tests to interface with postgres. Sqlalchemy provides ORM, multi database connection capability, and a lot of other functionality that we probably don't need in a test suite for a postgres extension. Further it may make it more difficult for downstream package managers to run our tests.


imor:
@AndrewJackson2020 this PR introduced  psycopg but did not remove sqlalchemy, what's the intention? Use both or remove sqlalchemy completely?

AndrewJackson2020:
> @AndrewJackson2020 this PR introduced psycopg but did not remove sqlalchemy, what's the intention? Use both or remove sqlalchemy completely?

Apologies, the initial comment was a bit underspecified, I think I discussed more details on slack.

The intention here is to provide a basic infrastructure so that we can start using psycopg directly in tests. I figure we could do something like:
1. provide basic test infra for psycopg
2. write all new tests with psycopg, no new tests are written in sqlalchemy
3. any PR that modifies an existing sqlalchemy test should rewrite the test to psycopg
4. move tests over in a one off manner
5. mass movement off all existing tests to psycopg (probably with AI), remove sqlalchemy from external dependencies

This was my idea but I'm open to being more aggressive (one time full AI rewrite?) if that is what we want to do.

imor:
Ok, got it. Up to you how you want to handle this. Smaller PRs are definitely easier to review. Stacked PRs work great for such use cases.

AndrewJackson2020:
> Ok, got it. Up to you how you want to handle this. Smaller PRs are definitely easier to review. Stacked PRs work great for such use cases.

My preference would be treating this as a long term directional goal using the 5 step process mentioned above and not do it all at once, at least for now.

I've added some language in the CONTRIBUTING.md explaining not to use SQLAlchemy for new tests. I've also outlined this process in the commit message. Can you please review and let me know if you are cool with this approach and thus approving the PR?

imor (approved):
Sounds good @AndrewJackson2020

PR #294, opened 2026-09-16
Python formatting

AndrewJackson2020:
Currently we do not have any formatting over the python codebase in `pg_net`. This PR adds ruff formatting and formats all code accordingly.

This PR consists of 2 commits:
commit 1: adds all necessary infra for ruff formatting
commit 2: runs ruff formatting (all changes were made by ruff, none by me directly)


steve-chavez:
Both these commits should be prefixed with `chore:`. To enforce that we can add commitlint on CI

AndrewJackson2020:
> Both these commits should be prefixed with `chore:`. To enforce that we can add commit lint on CI

Will fix this.

Also +1 for commit lint. I can add that as well.

AndrewJackson2020:
> Both these commits should be prefixed with `chore:`.

This has been fixed.

AndrewJackson2020:
It looks like there are some stuff that can't be auto formatted that need to be either fixes or ignored. I feel like its safest to make exceptions to stuff like that for now and reassess later if we want to make formatting more strict.

AndrewJackson2020:
Okay ruff checks are all working now, ended up making some unsafe fixes that looked reasonable and one manual fix. All of these steps are separate commits for traceability.

imor (approved):
LGTM

PR #295, opened 2026-09-17
chore: add conventional commit checker to GHA

AndrewJackson2020:
My last few contributions deviated from the standard of using conventional commits. Encoding that into CI since that is the most reliable way to ensure that doesn't get missed in review in the future.


brainrake (approved):

brainrake:
It works 🎉  https://github.com/supabase/pg_net/actions/runs/35177811911/job/105065449753#step:6:16

